Issue #4, opened 2018-05-20
Use crosshair cursor


emersion:
Probably w/ wayland-cursor

PR #8, opened 2018-06-25
Display dimensions of selection

Limero:
This will show the dimensions of current selection if slurp is started with the -d option.

slurp

Two things that I'm unsure of:
1. Is sprintf the best way to make a string from the coordinates?
2. Does anything else has to be done for HiDPI? I don't have a monitor to test with.

Note that the values (15 and 25) in `cairo_move_to(cairo, x + width + 15, y + height + 25);` are temporary and will have to be tweaked properly once #4 is done.

There are probably more things wrong with this code, so feel free to point out anything that can be improved :)

### DIFF
--- a/render.c
+++ b/render.c
@@ -19,6 +19,7 @@ void render(struct slurp_output *output) {
 	struct pool_buffer *buffer = output->current_buffer;
 	cairo_t *cairo = buffer->cairo;
 	int32_t scale = output->scale;
+	bool display_dimensions = output->display_dimensions;
 
 	uint32_t border_color = 0x000000FF;
 	int border_size = 2;
@@ -39,6 +40,17 @@ void render(struct slurp_output *output) {
 
 		int x, y, width, height;
 		pointer_get_box(pointer, &x, &y, &width, &height);
+
+		if (display_dimensions) {
+			cairo_select_font_face(cairo, "Sans", CAIRO_FONT_SLANT_NORMAL,
+				CAIRO_FONT_WEIGHT_NORMAL);
+			cairo_set_font_size(cairo, 14);
+			// buffer of 11 can hold selections up to 99999x99999
+			char dimensions[11];
+			sprintf(dimensions, "%ix%i", width, height);
+			cairo_move_to(cairo, x + width + 15, y + height + 25);
+			cairo_show_text(cairo, dimensions);
+		}
 
 		// Draw border
 		set_source_u32(cairo, border_color);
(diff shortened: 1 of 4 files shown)


render.c
```diff
@@ -40,6 +41,17 @@ void render(struct slurp_output *output) {
 		int x, y, width, height;
 		pointer_get_box(pointer, &x, &y, &width, &height);
 
+		if (display_dimensions) {
+			cairo_select_font_face(cairo, "Sans", CAIRO_FONT_SLANT_NORMAL,
+				CAIRO_FONT_WEIGHT_NORMAL);
+			cairo_set_font_size(cairo, 14);
+			// buffer of 11 can hold selections up to 99999x99999
+			char dimensions[11];
```
emersion: Need space for the final \0 too, but w/e

render.c
```diff
@@ -40,6 +41,17 @@ void render(struct slurp_output *output) {
...
+			sprintf(dimensions, "%ix%i", width, height);
```
emersion: Can you use `snprintf` with `sizeof(dimensions)`, just in case?

render.c
```diff
@@ -40,6 +41,17 @@ void render(struct slurp_output *output) {
...
+			cairo_move_to(cairo, x + width + 15, y + height + 25);
+			cairo_show_text(cairo, dimensions);
```
emersion: Hmm, I'm afraid we're going to need more elaborate stuff for HiDPI. Probably with `pango_attr_list_insert(attrs, pango_attr_scale_new(scale));`, see https://github.com/emersion/mako/blob/master/render.c#L76 (though we don't need the pango markup stuff).

To test HiDPI you can compile wlroots, configure the `WL-1` output with `scale=2`, run rootston inside another Wayland/X11 session and start slurp from within rootston. Not that easy, but doable ;)

SirCmpwn:
Another useful way to test hidpi is to run sway on the wayland backend and set the scale=2.

emersion:
Can you rebase you work on top of master please?

Limero:
Yes, I will. Should I move the "display_dimensions" variable to the same "slurp_state state" struct used for the other options created by @SirCmpwn?

emersion:
>Yes, I will. 

Sweet.

> Should I move the "display_dimensions" variable to the same "slurp_state state" struct used for the other options created by @SirCmpwn?

Yes, that's probably a good idea.

Limero:
It should all be fixed now. Is it possible for this to get merged and to create a separate issue (for me or anyone else) to add HiDPI support later?

SirCmpwn:
Were you unable to get a hidpi test setup working with the suggested approach?

Limero:
I just took another look at it and it was really easy with your sway approach. So I did another commit with HiDPI fixes :)
`swaymsg -t get_outputs`
`sway output eDP-1 scale 2`

PR #10, opened 2018-08-16
Track seats to free them on exit

ianyfan:
Fixes memory leaks. Just added a struct to track the seats, don't know if you want to structure it differently.

Also includes a small fix to remove destroyed outputs from list.


emersion:
Thanks!

PR #25, opened 2019-02-12
Abort on escape

yorickvP:
This looked doable. Is any more bookkeeping needed? maybe a `wl_keyboard_destroy` somewhere?

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -22,7 +22,6 @@ struct slurp_state {
 	struct wl_compositor *compositor;
 	struct zwlr_layer_shell_v1 *layer_shell;
 	struct wl_list outputs; // slurp_output::link
-	struct wl_list pointers; // slurp_pointer::link
 	struct wl_list seats; // slurp_seat::link
 
 	struct {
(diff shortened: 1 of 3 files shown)


emersion:
>maybe a `wl_keyboard_destroy` somewhere?

Yeah, we need to keep track of keyboards, just like pointers and seats.

I think it would make sense to merge `slurp_pointer` and `slurp_seat` together, since there is always at most one pointer per seat. Then we could add the keyboard state to the seat as well.

include/slurp.h
```diff
@@ -59,25 +58,23 @@ struct slurp_output {
...
-struct slurp_seat {
-	struct wl_seat *wl_seat;
-	struct wl_list link; // slurp_state::seats
+	int32_t x, y;
+	int32_t pressed_x, pressed_y;
 };
 
-void pointer_get_box(struct slurp_pointer *pointer, int *x, int *y,
-	int *width, int *height);
+void seat_get_box(struct slurp_seat *seat, struct slurp_box * result);
```
emersion: Style: `struct slurp_box *result`

main.c
```diff
@@ -98,47 +97,48 @@ static const struct wl_pointer_listener pointer_listener = {
...
+	result->x = min(seat->pressed_x, seat->x);
+	result->y = min(seat->pressed_y, seat->y);
+	result->width = abs(seat->x - seat->pressed_x);
+	result->height = abs(seat->y - seat->pressed_y);
 }
 
-static int min(int a, int b) {
-	return (a < b) ? a : b;
+static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
+	uint32_t serial, uint32_t time, uint32_t key, uint32_t key_state) {
```
emersion: Style: we indent continuation lines for function blocks with two tabs, to tell the different between the arguments and the body.

main.c
```diff
@@ -152,13 +152,23 @@ static void create_seat(struct slurp_state *state, struct wl_seat *wl_seat) {
 		fprintf(stderr, "allocation failed\n");
 		return;
 	}
+	seat->state = state;
+	seat->wl_pointer = NULL;
+	seat->wl_keyboard = NULL;
```
emersion: No need for these NULL assignments since we use `calloc`

emersion:
Thanks!

PR #27, opened 2019-02-12
Support selection across multiple outputs

yorickvP:
This builds on #25, sorry.

Adds support for selection across multiple outputs.

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -7,6 +7,7 @@
 
 #include "pool-buffer.h"
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
+#include "xdg-output-unstable-v1-client-protocol.h"
 
 struct slurp_box {
 	int32_t x, y;
@@ -21,6 +22,7 @@ struct slurp_state {
 	struct wl_shm *shm;
 	struct wl_compositor *compositor;
 	struct zwlr_layer_shell_v1 *layer_shell;
+	struct zxdg_output_manager_v1 *xdg_output_manager;
 	struct wl_list outputs; // slurp_output::link
 	struct wl_list seats; // slurp_seat::link
 
@@ -46,6 +48,8 @@ struct slurp_output {
 
 	struct wl_surface *surface;
 	struct zwlr_layer_surface_v1 *layer_surface;
+
+	struct zxdg_output_v1 *xdg_output;
 
 	struct wl_callback *frame_callback;
 	bool configured;
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -80,5 +80,5 @@ struct slurp_seat {
 };
 
 void seat_get_box(struct slurp_seat *seat, struct slurp_box *result);
-bool box_intersect(struct slurp_box *a, struct slurp_box *b);
+bool box_intersect(const struct slurp_box *a, const struct slurp_box *b);
 #endif
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -43,8 +43,7 @@ struct slurp_output {
 	struct slurp_state *state;
 	struct wl_list link; // slurp_state::outputs
 
-	struct slurp_box raw_geometry;
-	struct slurp_box geometry; // scaled
+	struct slurp_box geometry;
 	struct slurp_box logical_geometry;
 	int32_t scale;
 
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -43,7 +43,8 @@ struct slurp_output {
 	struct slurp_state *state;
 	struct wl_list link; // slurp_state::outputs
 
-	struct slurp_box geometry;
+	struct slurp_box raw_geometry;
+	struct slurp_box geometry; // scaled
 	struct slurp_box logical_geometry;
 	int32_t scale;
 
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -44,6 +44,7 @@ struct slurp_output {
 	struct wl_list link; // slurp_state::outputs
 
 	struct slurp_box geometry;
+	struct slurp_box logical_geometry;
 	int32_t scale;
 
 	struct wl_surface *surface;
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -76,5 +76,5 @@ struct slurp_seat {
 };
 
 void seat_get_box(struct slurp_seat *seat, struct slurp_box *result);
-
+bool box_intersect(struct slurp_box *a, struct slurp_box *b);
 #endif
--- a/main.c
+++ b/main.c
@@ -17,12 +17,13 @@ static void noop() {
 
 static void set_output_dirty(struct slurp_output *output);
 
-bool box_intersect(struct slurp_box *a, struct slurp_box *b) {
+bool box_intersect(const struct slurp_box *a, const struct slurp_box *b) {
 	return (a->x < b->x + b->width &&
-					a->x + a->width > b->x &&
-					a->y < b->y + b->height &&
-					a->height + a->y > b->y);
+	        a->x + a->width > b->x &&
+	        a->y < b->y + b->height &&
+	        a->height + a->y > b->y);
 }
+
 static struct slurp_output *output_from_surface(struct slurp_state *state,
 	struct wl_surface *surface);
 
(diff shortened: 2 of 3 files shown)


emersion:
Unfortunately using the output mode isn't going to work with fractional scaling. We'll need to use xdg-output to get the logical size and position. You can have a look at `grim` to see how this can be done.

emersion:
Can you rebase this PR?

include/slurp.h
```diff
@@ -76,5 +80,5 @@ struct slurp_seat {
 };
 
 void seat_get_box(struct slurp_seat *seat, struct slurp_box *result);
-
+bool box_intersect(struct slurp_box *a, struct slurp_box *b);
```
emersion: Arguments can be `const` here

main.c
```diff
@@ -17,6 +17,12 @@ static void noop() {
 
 static void set_output_dirty(struct slurp_output *output);
 
+bool box_intersect(struct slurp_box *a, struct slurp_box *b) {
+	return (a->x < b->x + b->width &&
+					a->x + a->width > b->x &&
```
emersion: Style: indent with one tab

main.c
```diff
@@ -52,17 +58,30 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 	seat->current_output = NULL;
 }
 
+static void seat_set_outputs_dirty(struct slurp_seat* seat) {
+	if (seat->button_state == WL_POINTER_BUTTON_STATE_PRESSED) {
+		struct slurp_box box;
+		seat_get_box(seat, &box);
+		struct slurp_output *output;
+		wl_list_for_each(output, &seat->state->outputs, link) {
+			if(box_intersect(&output->geometry, &box)) {
```
emersion: Style: missing space after `if`

main.c
```diff
@@ -52,17 +58,30 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 	seat->current_output = NULL;
 }
 
+static void seat_set_outputs_dirty(struct slurp_seat* seat) {
+	if (seat->button_state == WL_POINTER_BUTTON_STATE_PRESSED) {
```
emersion: Style: indentation in this function can be decreased with an early `return`

main.c
```diff
@@ -52,17 +58,30 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 	seat->current_output = NULL;
 }
 
+static void seat_set_outputs_dirty(struct slurp_seat* seat) {
```
emersion: Style: `*` on the right

main.c
```diff
@@ -181,20 +196,54 @@ static void output_handle_geometry(void *data, struct wl_output *wl_output,
 	output->geometry.y = y;
 }
 
+static void output_handle_mode(void *data, struct wl_output *wl_output,
+															 uint32_t flags, int32_t width, int32_t height, int32_t refresh) {
```
emersion: Style: indentation is a little messed up here.

main.c
```diff
@@ -207,7 +256,12 @@ static void create_output(struct slurp_state *state,
 	output->scale = 1;
 	wl_list_insert(&state->outputs, &output->link);
 
-	wl_output_add_listener(wl_output, &output_listener, output);
+	if (state->xdg_output_manager) {
```
emersion: This won't do, because the xdg-output-manager global can be advertised after all outputs. We need to create xdg-outputs and add listeners after the first roundtrip in `main` (and roundtrip a second time to retrieve xdg-output properties).

render.c
```diff
@@ -28,13 +28,15 @@ void render(struct slurp_output *output) {
...
 		if (!seat->wl_pointer) continue;
-		if (seat->button_state != WL_POINTER_BUTTON_STATE_PRESSED ||
-				seat->current_output != output) {
+		if (seat->button_state != WL_POINTER_BUTTON_STATE_PRESSED) {
 			continue;
 		}
 
 		struct slurp_box b;
 		seat_get_box(seat, &b);
+		if(!box_intersect(&output->geometry, &b)) continue;
```
emersion: Style: always use brackets, missing space after `if`

main.c
ianyfan: This check is still needed for hotplugged outputs though.

emersion: Hmm, I don't think we support these at all. because we don't create layer-shell surfaces for those. So this would be a follow-up issue.

render.c
```diff
@@ -28,13 +28,17 @@ void render(struct slurp_output *output) {
...
-		if (seat->button_state != WL_POINTER_BUTTON_STATE_PRESSED ||
-				seat->current_output != output) {
+		if (seat->button_state != WL_POINTER_BUTTON_STATE_PRESSED) {
 			continue;
 		}
 
 		struct slurp_box b;
 		seat_get_box(seat, &b);
+		if (!box_intersect(&output->logical_geometry, &b)) {
+      continue;
```
emersion: Indentation is a little messed up here

main.c
```diff
@@ -494,6 +561,7 @@ int main(int argc, char *argv[]) {
 		}
 		output->cursor_image = cursor->images[0];
 	}
+	wl_display_roundtrip(state.display);
```
emersion: Can we add a comment "Second rountrip for xdg-output" here?

main.c
```diff
@@ -464,7 +523,15 @@ int main(int argc, char *argv[]) {
 			state.layer_shell, output->surface, output->wl_output,
 			ZWLR_LAYER_SHELL_V1_LAYER_OVERLAY, "selection");
 		zwlr_layer_surface_v1_add_listener(output->layer_surface,
-			&layer_surface_listener, output);
+		  &layer_surface_listener, output);
+
+		if (state.xdg_output_manager) {
+			output->xdg_output = zxdg_output_manager_v1_get_xdg_output(state.xdg_output_manager, output->wl_output);
```
emersion: Style: can you split this line, it's longer than 80 chars

main.c
```diff
@@ -343,6 +397,8 @@ static void handle_global(void *data, struct wl_registry *registry,
 		struct wl_output *wl_output =
 			wl_registry_bind(registry, name, &wl_output_interface, 3);
 		create_output(state, wl_output);
+	} else if (strcmp(interface, zxdg_output_manager_v1_interface.name) == 0) {
+		state->xdg_output_manager = wl_registry_bind(registry, name, &zxdg_output_manager_v1_interface, (version > 2 ? 2 : version));
```
emersion: We just need xdg-output v1 here

main.c
```diff
@@ -181,20 +198,54 @@ static void output_handle_geometry(void *data, struct wl_output *wl_output,
 	output->geometry.y = y;
 }
 
+static void output_handle_mode(void *data, struct wl_output *wl_output,
+		uint32_t flags, int32_t width, int32_t height, int32_t refresh) {
+	struct slurp_output *output = data;
+
+	output->geometry.width = width;
```
emersion: We should only do this if it's the current mode.

main.c
```diff
@@ -181,20 +198,54 @@ static void output_handle_geometry(void *data, struct wl_output *wl_output,
...
 static void output_handle_scale(void *data, struct wl_output *wl_output,
 		int32_t scale) {
 	struct slurp_output *output = data;
 
 	output->scale = scale;
 }
+static void output_handle_done(void *data, struct wl_output *wl_output) {
+	struct slurp_output *output = data;
+	output->geometry.width /= output->scale;
+	output->geometry.height /= output->scale;
```
emersion: This won't work if the compositor sends this sequence:

```
mode
scale
done
scale
done
```

render.c
```diff
@@ -28,13 +28,17 @@ void render(struct slurp_output *output) {
...
-				seat->current_output != output) {
+		if (seat->button_state != WL_POINTER_BUTTON_STATE_PRESSED) {
 			continue;
 		}
 
 		struct slurp_box b;
 		seat_get_box(seat, &b);
+		if (!box_intersect(&output->logical_geometry, &b)) {
+			continue;
+    }
```
emersion: This line has an indentation issue too.

main.c
```diff
@@ -187,14 +214,41 @@ static void output_handle_scale(void *data, struct wl_output *wl_output,
 
 	output->scale = scale;
 }
+static void output_handle_done(void *data, struct wl_output *wl_output) {
+	struct slurp_output *output = data;
+	output->geometry = output->raw_geometry;
```
emersion: Err, this is quite a lot of complexity. Instead, can we just keep only `geometry` and `logical_geometry`, add the scale as a new field, and divide by the scale in `main` after `output->logical_geometry = output->geometry`?

main.c
```diff
@@ -52,17 +59,31 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
...
+		if (box_intersect(&output->logical_geometry, &box)) {
+			set_output_dirty(output);
+		}
+	}
+}
+
 static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
 	struct slurp_seat *seat = data;
+	// the places the cursor moved away from are also dirty
```
emersion: We should probably only do all of this if `seat->button_state == WL_POINTER_BUTTON_STATE_PRESSED`?

main.c
```diff
@@ -28,13 +35,13 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
...
-	seat->x = wl_fixed_to_int(surface_x);
-	seat->y = wl_fixed_to_int(surface_y);
-
 	// TODO: handle multiple overlapping outputs
 	seat->current_output = output;
 
+	seat->x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
+	seat->y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+
+
```
emersion: Extra newline here

main.c
yorickvP: Storing x and y is important if the button is not pressed.

render.c
yorickvP: Dammit. Really fighting emacs here.

emersion:
Thanks!

main.c
emersion: Follow-up issue: https://github.com/emersion/slurp/issues/32

PR #30, opened 2019-02-20
Display alpha in usage section

zeroeightysix:
`slurp -h` displays `#rrggbb` for colours, while slurp accepts `#rrggbbaa`.


emersion:
Thanks!

PR #31, opened 2019-02-20
Allow downstream to disable manpages even if scdoc is found

jbeich:
Similar to https://github.com/swaywm/sway/pull/3452.


emersion:
Thanks!

PR #33, opened 2019-02-20
Add user-defined output formatting

### LINKED ISSUE #1: Custom output format
`-f "%x,%y %wx%h"`

Might also want to add `%o` for the output name.

zeroeightysix:
Closes #1.
I hope my code is up to standards. I've never touched C before.

### DIFF
--- a/main.c
+++ b/main.c
@@ -435,7 +435,7 @@ uint32_t parse_color(const char *color) {
 }
 
 static void format_result(const char *format, char *buffer, const int buflen, struct slurp_box result) {
-	int bufferPos = 0;
+	int buffer_pos = 0;
 	for (int i = 0; format[i] != 0; i++) {
 		char c = format[i];
 		if (c == '%') {
@@ -458,15 +458,15 @@ static void format_result(const char *format, char *buffer, const int buflen, st
 			}
 
 			if (replacement != NULL) {
-				strncpy(buffer + bufferPos, replacement, buflen - bufferPos - 1);
-				bufferPos = bufferPos + strlen(replacement);
+				strncpy(buffer + buffer_pos, replacement, buflen - buffer_pos - 1);
+				buffer_pos = buffer_pos + strlen(replacement);
 				i++;
 				continue;
 			}
 		}
 
-		buffer[bufferPos] = c;
-		bufferPos++;
+		buffer[buffer_pos] = c;
+		buffer_pos++;
 	}
 	buffer[buflen - 1] = '\0';
 }
--- a/main.c
+++ b/main.c
@@ -434,27 +434,27 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
-static void format_result(const char *format, char *buffer, const int buflen, struct slurp_box result) {
+static void format_result(const struct slurp_box result, const char *format, char *buffer, const int buflen) {
 	int buffer_pos = 0;
-	for (int i = 0; format[i] != 0; i++) {
+	for (size_t i = 0; format[i] != 0; i++) {
 		char c = format[i];
 		if (c == '%') {
 			char next = format[i + 1];
 			
 			char replacement[128] = {0};
 			switch (next) {
-				case 'x':
-					sprintf(replacement, "%u", result.x);
-					break;
-				case 'y':
-					sprintf(replacement, "%u", result.y);
-					break;
-				case 'w':
-					sprintf(replacement, "%u", result.width);
-					break;
-				case 'h':
-					sprintf(replacement, "%u", result.height);
-					break;
+			case 'x':
+				sprintf(replacement, "%u", result.x);
+				break;
+			case 'y':
+				sprintf(replacement, "%u", result.y);
+				break;
+			case 'w':
+				sprintf(replacement, "%u", result.width);
+				break;
+			case 'h':
+				sprintf(replacement, "%u", result.height);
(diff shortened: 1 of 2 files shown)


slurp.1.scd
```diff
@@ -34,10 +34,25 @@ select, or click to cancel the selection.
...
 	Set border weight.
 
+*-f* _format_
+	Set format. See *FORMAT* for more detail.
+
 # COLORS
 
 Colors may be specified in #RRGGBB or #RRGGBBAA format. The # is optional.
 
+# FORMAT
```
emersion: Maybe this could say what the default is.

main.c
```diff
@@ -411,7 +411,10 @@ static const char usage[] =
 	"  -b #rrggbbaa Set background color.\n"
 	"  -c #rrggbbaa Set border color.\n"
 	"  -s #rrggbbaa Set selection color.\n"
-	"  -w n         Set border weight.\n";
+	"  -w n         Set border weight.\n"
+	"  -f s         Set output format.\n";
+
+static char format[128] = "%x,%y %wx%h";
```
emersion: This should be a local variable in `main`, and can be `strdup`'ed so that we don't have a length limit.

main.c
```diff
@@ -597,7 +640,8 @@ int main(int argc, char *argv[]) {
 		return EXIT_FAILURE;
 	}
 
-	printf("%d,%d %dx%d\n", state.result.x, state.result.y,
-		state.result.width, state.result.height);
+	char output_result[256] = {0};
```
emersion: Instead of using a buffer, we could just `printf` the string directly. This would save us from having to deal with buffer overflows and buffer position.

main.c
```diff
@@ -431,6 +434,43 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
+static void format_result(const char *format, char *buffer, const int buflen, struct slurp_box result) {
+	int bufferPos = 0;
```
emersion: Style: use `snake_case`

main.c
```diff
@@ -431,6 +434,43 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
+static void format_result(const char *format, char *buffer, const int buflen, struct slurp_box result) {
```
emersion: I'd prefer the first argument to be `result`, as the function is named `format_result`. Also, `result` could be a `const struct slurp_box *`.

main.c
```diff
@@ -431,6 +434,43 @@ uint32_t parse_color(const char *color) {
...
+	int bufferPos = 0;
+	for (int i = 0; format[i] != 0; i++) {
+		char c = format[i];
+		if (c == '%') {
+			char next = format[i + 1];
+			
+			char replacement[128] = {0};
+			switch (next) {
+				case 'x':
```
emersion: Style: we don't indent `case` statements (`case`s have the same indentation level as `switch`)

main.c
```diff
@@ -431,6 +434,43 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
+static void format_result(const char *format, char *buffer, const int buflen, struct slurp_box result) {
+	int bufferPos = 0;
+	for (int i = 0; format[i] != 0; i++) {
```
emersion: `i` can be `size_t`

main.c
zeroeightysix: Not sure why you'd use `strdup` here. Instead, I chose to initialise `format` to `NULL`, point it to `optarg` if the format is specified, and if not, set it to the default format.

emersion: Oh, right, that should work :+1:

main.c
```diff
@@ -431,6 +432,34 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
+static void print_formatted_result(const struct slurp_box result, const char *format) {
```
emersion: Can `result` be a pointer instead?

main.c
```diff
@@ -431,6 +432,34 @@ uint32_t parse_color(const char *color) {
...
+	for (size_t i = 0; format[i] != 0; i++) {
+		char c = format[i];
+		if (c == '%') {
+			char next = format[i + 1];
+			
+			i++; // Skip the next character (x, y, w or h)
+			switch (next) {
+			case 'x':
+				printf("%u", result.x);
+				continue;
```
emersion: `continue` is a little bit weird here. Can we use `break` instead?

emersion: Ah, I see. `continue` won't affect the `for` loop, since it's wrapped in a `switch`. Probably need to use a `default` case.

main.c
```diff
@@ -476,6 +509,10 @@ int main(int argc, char *argv[]) {
 		}
 	}
 
+	if (format == NULL) {
+		format = "%x,%y %wx%h";
```
emersion: Could just use this as the initialization value

main.c
zeroeightysix: We're continuing to keep the `i--` from happening and the character from being printed. I could break and put `printf("%c", c)` in an else-block, remove 'i--` and put `i++` in the case blocks instead of before the switch, but that's just more lines and repetition, isn't it?

emersion: Eh, I see that `continue` does affect the loop instead of the `switch`. It's pretty confusing, but I guess it makes sense.

I'm fine with this approach.

emersion:
Thanks!

PR #28, opened 2019-02-14
Accept list of predefined regions on stdin for quick selection

yorickvP:
Sorry for building on #27. I can probably rebase if neccesary. Actual compare view is at https://github.com/yorickvP/slurp/compare/multi-outputs...yorickvP:select-predef

This accepts a list of regions on stdin (if stdin is non-interactive),
i.e. `echo -e "2565,31 1855x1404\n4425,31 690x1404" | slurp`

Dragging still works, but a single click will select the hovered region. This can be used to implement some features:

#18: `swaymsg -t get_outputs | jq -r '.[] | select(.active) | .rect | "\(.x),\(.y) \(.width)x\(.height)"' | slurp`
#16: `swaymsg -t get_tree | jq -r '.. | (.nodes? // empty)[] | select(.pid and .visible) | .rect | "\(.x),\(.y) \(.width)x\(.height)"'  | slurp`

I need to figure out a jq that filters visible windows. Not sure if there are any other usecases. This maybe should just be a wayland protocol.

PR #35, opened 2019-04-07
Move selection on `space` keypress

LeonardDrs:
Hi,

This is a feature add.

When dragging the selection, you can  use the modifier `space` key to move your selection around 

Video:
output
Result on second screen (not from the same recording and my screens aren't aligned):
output2

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -17,6 +17,7 @@ struct slurp_box {
 
 struct slurp_state {
 	bool running;
+	bool edit_anchor;
 
 	struct wl_display *display;
 	struct wl_registry *registry;
@@ -80,7 +81,7 @@ struct slurp_seat {
 	enum wl_pointer_button_state button_state;
 	struct slurp_output *current_output;
 	int32_t x, y;
-	int32_t pressed_x, pressed_y;
+	int32_t anchor_x, anchor_y;
 	struct slurp_box selection;
 	bool has_selection;
 };
--- a/main.c
+++ b/main.c
@@ -14,7 +14,6 @@ static void noop() {
 	// This space intentionally left blank
 }
 
-
 static void set_output_dirty(struct slurp_output *output);
 
 bool box_intersect(const struct slurp_box *a, const struct slurp_box *b) {
@@ -27,6 +26,20 @@ bool box_intersect(const struct slurp_box *a, const struct slurp_box *b) {
 static struct slurp_output *output_from_surface(struct slurp_state *state,
 	struct wl_surface *surface);
 
+static void move_pointer(struct slurp_seat *seat, wl_fixed_t surface_x,
+		wl_fixed_t surface_y) {
+	int x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
+	int y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+
+	if (seat->state->edit_anchor) {
+		seat->anchor_x += x - seat->x;
+		seat->anchor_y += y - seat->y;
+	}
+
+	seat->x = x;
+	seat->y = y;
+}
+
 static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 		uint32_t serial, struct wl_surface *surface,
 		wl_fixed_t surface_x, wl_fixed_t surface_y) {
@@ -38,8 +51,7 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 	// TODO: handle multiple overlapping outputs
 	seat->current_output = output;
 
-	seat->x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
-	seat->y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+	move_pointer(seat, surface_x, surface_y);
 
 	wl_surface_set_buffer_scale(seat->cursor_surface, output->scale);
 	wl_surface_attach(seat->cursor_surface,
@@ -82,6 +94,10 @@ static int min(int a, int b) {
 	return (a < b) ? a : b;
 }
 
+static int max(int a, int b) {
+	return (a > b) ? a : b;
+}
+
 static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
 	struct slurp_seat *seat = data;
@@ -90,8 +106,7 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		seat_set_outputs_dirty(seat);
 	}
 
-	seat->x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
-	seat->y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+	move_pointer(seat, surface_x, surface_y);
 
 	switch (seat->button_state) {
 	case WL_POINTER_BUTTON_STATE_RELEASED:
@@ -110,12 +125,14 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 			}
 		}
 		break;
-	case WL_POINTER_BUTTON_STATE_PRESSED:
+	case WL_POINTER_BUTTON_STATE_PRESSED:;
+		int32_t anchor_x = max(0, seat->anchor_x);
+		int32_t anchor_y = max(0, seat->anchor_y);
 		seat->has_selection = true;
-		seat->selection.x = min(seat->pressed_x, seat->x);
-		seat->selection.y = min(seat->pressed_y, seat->y);
-		seat->selection.width = abs(seat->x - seat->pressed_x);
-		seat->selection.height = abs(seat->y - seat->pressed_y);
+		seat->selection.x = min(anchor_x, seat->x);
+		seat->selection.y = min(anchor_y, seat->y);
+		seat->selection.width = abs(seat->x - anchor_x);
+		seat->selection.height = abs(seat->y - anchor_y);
 		break;
 	}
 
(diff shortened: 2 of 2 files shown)


main.c
```diff
@@ -132,6 +137,20 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
 		if (key == KEY_ESC) {
 			state->running = false;
 		}
+
+		if (key == KEY_SPACE) {
```
LeonardDrs: Note: Weird behavior if holding `space` before clicking.
The selection starts at the top left corner to the mouse position. Any mouse movement after that works like intended with or without releasing `space`.

Should we check for the selection before setting `edit_anchor`?

main.c
```diff
@@ -77,8 +84,16 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		seat_set_outputs_dirty(seat);
 	}
 
-	seat->x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
-	seat->y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+	int32_t x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
+	int32_t y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+
+	if (seat->state->edit_anchor) {
+		seat->anchor_x = max(0, seat->anchor_x + (x - seat->x));
```
LeonardDrs: I'm suspicious of this `0` here. 

Will it work with multiples monitors?
Will the selection be able to move freely between monitors? Should it?
Can't test at home, as I only have one.

LeonardDrs: After trying the Mac feature again, I realised I didn't implement the same behavior.
If you can test both commit behavior and tell me what you think is best, I would be a grateful.

emersion:
Can you explain what this feature is? I'm not familiar with macOS.

LeonardDrs:
Oh yeah my bad.

When taking a screenshot, you can select a part of the screen as well. During this selection, you can also use the modifier `space` key to move your selection around whilst the mouse button is pressed.

Video:
output
Result on second screen (not in sync):
output2


Also on mac, if you press `space` without holding the mouse button, it highlights the app you are currently hovering, and you can screen it with a click.
I left that part out since the PR #28 is around, and I have no clue on how to gather information on the app/panel coordinates under the cursor. I could give a try but I have no idea how.

LeonardDrs:
Hey @emersion!
Just saw that the PR I had conflict with was merged, that's nice :smile: 

I still have a couple of question, let me know if you need anything else!

main.c
```diff
@@ -171,8 +183,23 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
...
+			seat->origin_x = seat->x;
+			seat->origin_y = seat->y;
+			seat->origin_anchor_x = seat->anchor_x;
+			seat->origin_anchor_y = seat->anchor_y;
+		}
+	}
+
+	if (key_state == WL_KEYBOARD_KEY_STATE_RELEASED) {
+		if (key == KEY_SPACE) {
+			state->edit_anchor = false;
```
LeonardDrs: Is this check really necessary considering the one l.186?

main.c
```diff
@@ -110,13 +118,17 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
...
+			int32_t anchor_x = max(0, seat->anchor_x);
+			int32_t anchor_y = max(0, seat->anchor_y);
+			seat->has_selection = true;
+			seat->selection.x = min(anchor_x, seat->x);
+			seat->selection.y = min(anchor_y, seat->y);
+			seat->selection.width = abs(seat->x - anchor_x);
+			seat->selection.height = abs(seat->y - anchor_y);
+			break;
+
+	}
```
LeonardDrs: I have no idea how to indent this switch/case, tried my best regarding the rest of this file

main.c
```diff
@@ -110,13 +118,17 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
...
 		break;
-	case WL_POINTER_BUTTON_STATE_PRESSED:
-		seat->has_selection = true;
-		seat->selection.x = min(seat->pressed_x, seat->x);
-		seat->selection.y = min(seat->pressed_y, seat->y);
-		seat->selection.width = abs(seat->x - seat->pressed_x);
-		seat->selection.height = abs(seat->y - seat->pressed_y);
-		break;
+	case WL_POINTER_BUTTON_STATE_PRESSED: {
+			int32_t anchor_x = max(0, seat->anchor_x);
```
LeonardDrs: I'm doing this here and not [l.100](https://github.com/emersion/slurp/pull/35/files#diff-2045016cb90d1e65d71c2407a2570927R100) because we don't want the selection to be cropped if the user goes out of bounds

emersion:
Hmm, sorry for the delay. I'll have a look at this PR soon.

LeonardDrs:
Haha don't worry I'm just happy to contribute ☺️

main.c
emersion: We could check `button_state` to prevent this?

main.c
emersion: One indent level is enough. Instead of `case thing: { … }` you can do `case thing:; …` (note the semicolon) which doesn't mess up indentation.

main.c
emersion: You mean you want to prevent the selection to go out of bounds?

The same needs to be done for the upper bound, I suppose.

include/slurp.h
```diff
@@ -80,7 +81,9 @@ struct slurp_seat {
 	enum wl_pointer_button_state button_state;
 	struct slurp_output *current_output;
 	int32_t x, y;
-	int32_t pressed_x, pressed_y;
+	int32_t anchor_x, anchor_y;
+	int32_t origin_x, origin_y;
+	int32_t origin_anchor_x, origin_anchor_y;
```
emersion: Hmm, I'm not sure I understand the role of all these variables. What is origin? Do we really need all of these?

Instead, we could compute a delta before updating `state->x` and `state->y`:

```c
int x = /* new x value */;
int dx = state->x - x;
state->x = x;
// If we're editing the anchor, add dx
```

Thoughts?

main.c
LeonardDrs: I want the selection to go out of bound. I just don't want it to be cropped.
Example: I'm doing a 100x100 selection and I want to move it to the 0,0 position. If I go out of bounds atop of the screen by 10px, I don't want it to become a 100x90

main.c
```diff
@@ -117,11 +121,17 @@ static int min(int a, int b) {
...
+static int max(int a, int b) {
+	return (a > b) ? a : b;
+}
+
 void seat_get_box(struct slurp_seat *seat, struct slurp_box *result) {
-	result->x = min(seat->pressed_x, seat->x);
-	result->y = min(seat->pressed_y, seat->y);
-	result->width = abs(seat->x - seat->pressed_x);
-	result->height = abs(seat->y - seat->pressed_y);
+	int32_t anchor_x = max(0, seat->anchor_x);
```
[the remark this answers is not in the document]
LeonardDrs: If you release the selection while out of bound - because you want to start @0,0 for example - you don't want to select the negative part of your screen.

main.c
LeonardDrs: Yeah make sense, and ignore the fact that the user want to move the selection since there is no selection to move around. If the user wants to move it, he would have to release and press again.

Seems good :+1:

include/slurp.h
LeonardDrs: I'm pretty sure we can do it without all those variables but I couldn't find a way before.
I'll try with the delta. Thanks

main.c
LeonardDrs: Checked `seat->has_selection` instead. Felt cleaner.

main.c
LeonardDrs: An image is worth a thousand words:

No `max`:
output-no-max

`max`-ing when setting the `anchor`:
output-crop

LeonardDrs:
@emersion :scream: I got an issue with the following commit:
>Use delta instead of multiples variables

Somehow, if I move around my selection too fast between two monitors, it gets resized :scream: 
output
(End of first monitor on the left)

Tried moving fast on the same monitor, or out of bound and back, it's fine.

Any idea?

emersion:
Thanks for the update, the code looks much better now :D

I've tried to debug this, and after figuring out that the size of the box doesn't change between the beginning and the end of `pointer_handle_motion` (which is what we want), I figured out there is another place where we set `seat->x` and `y`: in `pointer_handle_enter`. This is the source of the bug: the function sets `seat->{x,y}` without updating the anchor.

I'd suggest adding a `move_pointer` function that updates the coords and the anchor. What do you think?

main.c
```diff
@@ -90,8 +93,16 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		seat_set_outputs_dirty(seat);
 	}
 
-	seat->x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
-	seat->y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+	int x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
+	int y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+
+	if (seat->state->edit_anchor) {
+		seat->anchor_x = seat->anchor_x + (x - seat->x);
```
emersion: This can be simplified to

```c
seat->anchor_x += x - seat->x;
```

emersion:
After these changes (and a rebase), this PR will be good to merge.

main.c
LeonardDrs: Yep forgot it when I was trying to debug :sweat_smile: Thanks

LeonardDrs:
Now that you gave me the answer, It seems pretty obvious.
I feel dummy, should have debugged better!

Thanks a lot :smile: 

`move_pointer` is neat indeed, hope I didn't made any style mistake...

LeonardDrs:
No issue whatsoever! I love using sway and its ecosystem :smile:

Issue #29, opened 2019-02-19
Touch support


emersion:

gg-rewrite:
Hello.
Just in case: I've created a PR #50 for this one. Can you please review it?

PR #50, opened 2020-03-05
#29: adding touch support 

gg-rewrite:
Selection by a single touch point, similar to the mouse.

Boilerplate code for touch handling is added.

Three pieces of code are moved to separate functions from handle_pointer* methods so as to be reused by the corresponding handle_touch* one.

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -9,6 +9,8 @@
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 #include "xdg-output-unstable-v1-client-protocol.h"
 
+#define TOUCH_ID_EMPTY -1
+
 struct slurp_box {
 	int32_t x, y;
 	int32_t width, height;
@@ -78,18 +80,24 @@ struct slurp_seat {
 	// keyboard:
 	struct wl_keyboard *wl_keyboard;
 
-	// pointer:
-	struct wl_pointer *wl_pointer;
-	enum wl_pointer_button_state button_state;
+	//selection (pointer/touch):
 	struct slurp_output *current_output;
 	int32_t x, y;
 	int32_t anchor_x, anchor_y;
 	struct slurp_box selection;
 	bool has_selection;
 
+	// pointer:
+	struct wl_pointer *wl_pointer;
+	enum wl_pointer_button_state button_state;
+
 	// keymap:
 	struct xkb_keymap *xkb_keymap;
 	struct xkb_state *xkb_state;
+	
+	// touch:
+	struct wl_touch *wl_touch;
+  	int32_t touch_id;
 };
 
 bool box_intersect(const struct slurp_box *a, const struct slurp_box *b);
--- a/main.c
+++ b/main.c
@@ -29,7 +29,7 @@ bool box_intersect(const struct slurp_box *a, const struct slurp_box *b) {
 static struct slurp_output *output_from_surface(struct slurp_state *state,
 	struct wl_surface *surface);
 
-static void move_pointer(struct slurp_seat *seat, wl_fixed_t surface_x,
+static void move_seat(struct slurp_seat *seat, wl_fixed_t surface_x,
 		wl_fixed_t surface_y) {
 	int x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
 	int y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
@@ -54,7 +54,7 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 	// TODO: handle multiple overlapping outputs
 	seat->current_output = output;
 
-	move_pointer(seat, surface_x, surface_y);
+	move_seat(seat, surface_x, surface_y);
 
 	wl_surface_set_buffer_scale(seat->cursor_surface, output->scale);
 	wl_surface_attach(seat->cursor_surface,
@@ -101,6 +101,17 @@ static int max(int a, int b) {
 	return (a > b) ? a : b;
 }
 
+static void handle_active_selection_motion(struct slurp_seat *seat) {
+	int32_t anchor_x = max(0, seat->anchor_x);
+	int32_t anchor_y = max(0, seat->anchor_y);
+	seat->has_selection = true;
+	seat->selection.x = min(anchor_x, seat->x);
+	seat->selection.y = min(anchor_y, seat->y);
+	// selection includes the seat and anchor positions
+	seat->selection.width = abs(seat->x - anchor_x) + 1;
+	seat->selection.height = abs(seat->y - anchor_y) + 1;
+}
+
 static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
 	struct slurp_seat *seat = data;
@@ -109,7 +120,7 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		seat_set_outputs_dirty(seat);
 	}
 
-	move_pointer(seat, surface_x, surface_y);
+	move_seat(seat, surface_x, surface_y);
 
 	switch (seat->button_state) {
 	case WL_POINTER_BUTTON_STATE_RELEASED:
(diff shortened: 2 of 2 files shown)


include/slurp.h
```diff
@@ -84,6 +84,11 @@ struct slurp_seat {
 	int32_t anchor_x, anchor_y;
 	struct slurp_box selection;
 	bool has_selection;
+
+	// touch:
+	struct wl_touch *wl_touch;
+	bool already_touched;
+	struct slurp_output *current_touch_output;
```
emersion: Do we really need this field? Can't we just use `current_output`?

include/slurp.h
```diff
@@ -84,6 +84,11 @@ struct slurp_seat {
 	int32_t anchor_x, anchor_y;
 	struct slurp_box selection;
 	bool has_selection;
+
+	// touch:
```
emersion: The fields used for both pointer and touch should be groupped into a separate section.

main.c
```diff
@@ -218,6 +231,49 @@ static const struct wl_keyboard_listener keyboard_listener = {
...
+}
+
+static const struct wl_touch_listener touch_listener = {
+	.down = touch_handle_down,
+	.up = touch_handle_up,
+	.frame = noop,
+	.motion = touch_handle_motion,
+	.orientation = noop,
+	.shape = noop,
+	.cancel = noop,
```
emersion: Should we clear our state when this event is received?

main.c
```diff
@@ -126,14 +137,7 @@ static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
...
 	case WL_POINTER_BUTTON_STATE_PRESSED:;
-		int32_t anchor_x = max(0, seat->anchor_x);
-		int32_t anchor_y = max(0, seat->anchor_y);
-		seat->has_selection = true;
-		seat->selection.x = min(anchor_x, seat->x);
-		seat->selection.y = min(anchor_y, seat->y);
-		// selection includes the seat and anchor positions
-		seat->selection.width = abs(seat->x - anchor_x) + 1;
-		seat->selection.height = abs(seat->y - anchor_y) + 1;
+		handle_active_selection_motion(seat);
```
emersion: Could maybe rename to `move_selection`?

main.c
```diff
@@ -218,6 +231,49 @@ static const struct wl_keyboard_listener keyboard_listener = {
...
+static void touch_handle_down(void *data, struct wl_touch *touch, uint32_t serial,
+                          uint32_t time, struct wl_surface *surface, int32_t id,
+                          wl_fixed_t x, wl_fixed_t y) {
+	struct slurp_seat *seat = data;
+	if (!seat->already_touched) {
+		seat->already_touched = true;
+		seat->current_output =
+			output_from_surface(seat->state, surface);
+		seat->current_touch_output = seat->current_output;
+		move_pointer(seat, x, y);
```
emersion: Since these coordinates are used for touch too, could rename to `move_seat`.

include/slurp.h
gg-rewrite: Yes but this one is a bit trickier.

I've investigated further and found out that
1. pointer_leave event is triggered whenever touch down happens (even if mouse is not moved).
2. handle_pointer_leave clears the current_output. 

I think the solution is either 
 * checking if has_selection is true and then leaving current_output be
 * leaving pointer_handle_leave empty since clearing current_output is the only thing the method does (until multi-monitor funcitonality is added).

Which one is better?

main.c
gg-rewrite: And done f2050fa6a7ca7c4cc72e122970ed00fdf21bc5bd.

main.c
gg-rewrite: Added the touch_cancel event fd789e9988ead866633f06917ccf339014714be9.

include/slurp.h
gg-rewrite: Here's the fix 6538ddf97a0822135dffe258ecfdb4112710c028

include/slurp.h
emersion: >pointer_leave event is triggered whenever touch down happens (even if mouse is not moved).

That sounds like a Sway bug.

gg-rewrite: > That sounds like a Sway bug.

Confirmed as this doesn't happen in wayfire. Opened an issue https://github.com/swaywm/sway/issues/5145 for this one.

emersion: Thanks for tracking it down!

emersion: The "real" solution would be to handle pointer input and touch input completely separately. Basically put the currently shared state in a struct and have two separate fields with this struct (one for pointer, one for touch). It would be a bigger change though.

gg-rewrite: Just to be sure I got it right:

```
//selection (pointer/touch):
	struct {
		struct slurp_output *current_output;
		struct slurp_output *current_touch_output;
		int32_t x, y;
		int32_t anchor_x, anchor_y;
		struct slurp_box selection;
		bool has_selection;
	} selection;
```

Something like this?

emersion: The idea is to track the position of the pointer and touch points independently. So we'd have:

```
struct slurp_selection {
	struct slurp_output *current_output;
	int32_t x, y;
	int32_t anchor_x, anchor_y;
	struct slurp_box box;
	bool valid;
};

struct slurp_seat {
	…

	struct slurp_selection pointer;
	struct slurp_selection touch;
};
```

However, I'm fine with keeping this for a future pull request.

Now that the Sway patch has been merged, we can get rid of `current_touch_output`.

gg-rewrite: And done 4e127e3386f9802c7976b07df5db2826efa36540.

main.c
```diff
@@ -218,6 +231,54 @@ static const struct wl_keyboard_listener keyboard_listener = {
 	.modifiers = noop,
 };
 
+static void touch_handle_down(void *data, struct wl_touch *touch, uint32_t serial,
+                          uint32_t time, struct wl_surface *surface, int32_t id,
```
emersion, line 235: Style: we don't align, we use two tabs instead.

main.c
```diff
@@ -218,6 +231,54 @@ static const struct wl_keyboard_listener keyboard_listener = {
...
+static void touch_handle_motion(void *data, struct wl_touch *touch, uint32_t time,
+                            int32_t id, wl_fixed_t x, wl_fixed_t y) {
+	struct slurp_seat *seat = data;
+	move_seat(seat, x, y);
+	handle_active_selection_motion(seat);
+	seat_set_outputs_dirty(seat);
+}
+
+static void touch_handle_cancel(void *data, struct wl_touch *touch)
+{
```
emersion, line 268: Style: no newline between `)` and `{`

main.c
```diff
@@ -218,6 +231,54 @@ static const struct wl_keyboard_listener keyboard_listener = {
 	.modifiers = noop,
 };
 
+static void touch_handle_down(void *data, struct wl_touch *touch, uint32_t serial,
+                          uint32_t time, struct wl_surface *surface, int32_t id,
+                          wl_fixed_t x, wl_fixed_t y) {
+	struct slurp_seat *seat = data;
+	if (!seat->already_touched) {
```
emersion, line 238: Instead of having a `bool already_touched` field, can we store the touch `id` and check it's the correct one in `touch_handle_motion`?

include/slurp.h
```diff
@@ -82,8 +84,8 @@ struct slurp_seat {
 	int32_t anchor_x, anchor_y;
 	struct slurp_box selection;
 	bool has_selection;
-	bool already_touched;
-	
+	int32_t touch_id;
```
emersion, line 87: One last nit:" can we move this to the "touch" section below?

main.c
gg-rewrite: Done 1e21b359fb7530f226e4c0e5d75f8d6a5befaabd.

main.c
gg-rewrite: Also done.

emersion:
If you could rebase and squash everything into a single commit, that'd make it ready for being merged :)

main.c
gg-rewrite: Fixed 585646d2ad2a322da034b52fa494d697584ed144. Turns out this also fixes segfaults when attempting to select with different fingers, thanks for this one.

emersion:
Can we pick a better commit summary + message? Something like:

```
Add touch support

Closes: https://github.com/emersion/slurp/issues/29
```

emersion:
Thanks!

PR #52, opened 2020-03-25
Take into account keymap

### LINKED ISSUE #36: slurp ignores keyboard layout
I got `XKB_DEFAULT_OPTIONS='compose:ralt,caps:swapescape'` in my environment (using sway).

When I invoke slurp and try to cancel it with the Caps key nothing happens.
When I press the actual Esc key it gets cancelled, but caps lock gets triggered as well.

NilsIrl:
Fix #36

The first commit contains debug code.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1,10 +1,13 @@
 #define _POSIX_C_SOURCE 2
+
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
 #include <unistd.h>
 #include <wayland-cursor.h>
+#include <xkbcommon/xkbcommon.h>
 #include <linux/input-event-codes.h>
 
 #include "slurp.h"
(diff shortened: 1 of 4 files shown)


main.c
```diff
@@ -182,36 +190,103 @@ static const struct wl_pointer_listener pointer_listener = {
...
 			state->edit_anchor = false;
 		}
+		break;
 	}
+
 }
 
 static const struct wl_keyboard_listener keyboard_listener = {
-	.keymap = noop,
+	.keymap = keyboard_handle_keymap,
```
emersion: Also need to listen to the `modifiers` event.

emersion:
How swaylock does it, for reference: https://github.com/swaywm/swaylock/blob/master/seat.c#L10

main.c
NilsIrl: That's what I thought at first, but aren't modifiers just CTRL, Mod.... which means that we don't really care about them (for the moment), as the only keys we care about are space and escape

emersion: The user may have configured his keyboard so that Shift+Caps Lock yields the Esc keysym, for instance.

main.c
```diff
@@ -1,10 +1,18 @@
 #define _POSIX_C_SOURCE 2
+
+#include <assert.h>
+#include <time.h>
+#include <sys/mman.h>
+#include <alloca.h>
```
emersion: This is unused.

main.c
```diff
@@ -1,10 +1,18 @@
...
+#include <err.h>
```
emersion: This is a non-standard API.

main.c
```diff
@@ -182,36 +190,66 @@ static const struct wl_pointer_listener pointer_listener = {
...
+static void keyboard_handle_keymap(void *data, struct wl_keyboard *wl_keyboard,
+		uint32_t format, int32_t fd, uint32_t size) {
+	struct slurp_seat *seat = data;
+	switch (format) {
+	case WL_KEYBOARD_KEYMAP_FORMAT_NO_KEYMAP:
+		seat->xkb_keymap = xkb_keymap_new_from_names(seat->state->xkb_context, NULL, XKB_KEYMAP_COMPILE_NO_FLAGS);
+		break;
+	case WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1:;
+		void *buffer;
+		if ((buffer = mmap(NULL, size - 1, PROT_READ, MAP_SHARED, fd, 0)) == MAP_FAILED) {
```
emersion: No need for this `- 1` which could underflow.

emersion: Prefer `MAP_PRIVATE` to `MAP_SHARED`.

main.c
```diff
@@ -623,6 +663,12 @@ int main(int argc, char *argv[]) {
 		return EXIT_FAILURE;
 	}
 
+	if ((state.xkb_context = xkb_context_new(XKB_CONTEXT_NO_FLAGS)) == NULL) {
+		// TODO for reviewer: Silent failure? or smth else?
```
emersion: "xkb_context_new failed" printed to stderr is fine

main.c
```diff
@@ -1,10 +1,18 @@
 #define _POSIX_C_SOURCE 2
+
+#include <assert.h>
+#include <time.h>
```
emersion: This is also unused.

main.c
NilsIrl: according to http://man7.org/linux/man-pages/man2/mmap.2.html, `MAP_PRIVATE` uses copy-on-write which (I suppose) has some overhead. Also we don't really need this feature since we're not 1. writing to that memory and 2. accessing that memory from other processes

emersion: See https://gitlab.freedesktop.org/wayland/wayland/-/blob/master/protocol/wayland.xml#L2182 for the reason why we should use `MAP_PRIVATE`

main.c
NilsIrl: addressed

main.c
NilsIrl: addressed

NilsIrl: If the mmap fails, now `exit(EXIT_FAILURE)` happens.

main.c
NilsIrl: addressed

main.c
NilsIrl: ok, I pushed a commit for that. Also, it returns 1 (i.e. exits)

main.c
NilsIrl: https://github.com/xkbcommon/libxkbcommon/blob/master/test/interactive-wayland.c#L341

it uses -1 but not for mmap (which is an inconsistency from the example). I guess the buffer should be a valid string so it must have a size of at least 1 (for the '\0').

Here it also uses the length without the '\0' https://github.com/xkbcommon/libxkbcommon/blob/d92a248c48227d09f6fdcfafaf339a5ff586e042/src/keymap.c#L164

main.c
NilsIrl: addressed

NilsIrl:
> Can you squash all of these commits into one

done

emersion:
Thanks for your contribution!

PR #55, opened 2020-04-11
separated pointer and touch selection states

gg-rewrite:
Added separated structs as per discussion in #50.

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -17,6 +17,14 @@ struct slurp_box {
 	struct wl_list link;
 };
 
+struct slurp_selection {
+	struct slurp_output *current_output;
+	int32_t x, y;
+	int32_t anchor_x, anchor_y;
+	struct slurp_box selection;
+	bool has_selection;
+};
+
 struct slurp_state {
 	bool running;
 	bool edit_anchor;
@@ -81,11 +89,9 @@ struct slurp_seat {
 	struct wl_keyboard *wl_keyboard;
 
 	//selection (pointer/touch):
-	struct slurp_output *current_output;
-	int32_t x, y;
-	int32_t anchor_x, anchor_y;
-	struct slurp_box selection;
-	bool has_selection;
+
+	struct slurp_selection pointer_selection;
+	struct slurp_selection touch_selection;
 
 	// pointer:
 	struct wl_pointer *wl_pointer;
--- a/main.c
+++ b/main.c
@@ -30,17 +30,19 @@ static struct slurp_output *output_from_surface(struct slurp_state *state,
 	struct wl_surface *surface);
 
 static void move_seat(struct slurp_seat *seat, wl_fixed_t surface_x,
-		wl_fixed_t surface_y) {
-	int x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
-	int y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+		wl_fixed_t surface_y,
+		struct slurp_selection *current_selection) {
+	int x = wl_fixed_to_int(surface_x) +
+		current_selection->current_output->logical_geometry.x;
+	int y = wl_fixed_to_int(surface_y) + current_selection->current_output->logical_geometry.y;
 
 	if (seat->state->edit_anchor) {
-		seat->anchor_x += x - seat->x;
-		seat->anchor_y += y - seat->y;
+		current_selection->anchor_x += x - current_selection->x;
+		current_selection->anchor_y += y - current_selection->y;
 	}
 
-	seat->x = x;
-	seat->y = y;
+	current_selection->x = x;
+	current_selection->y = y;
 }
 
 static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
@@ -52,9 +54,9 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 		return;
 	}
 	// TODO: handle multiple overlapping outputs
-	seat->current_output = output;
+	seat->pointer_selection.current_output = output;
 
-	move_seat(seat, surface_x, surface_y);
+	move_seat(seat, surface_x, surface_y, &seat->pointer_selection);
 
 	wl_surface_set_buffer_scale(seat->cursor_surface, output->scale);
 	wl_surface_attach(seat->cursor_surface,
@@ -70,13 +72,16 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 	struct slurp_seat *seat = data;
 
 	// TODO: handle multiple overlapping outputs
-	seat->current_output = NULL;
+	seat->pointer_selection.current_output = NULL;
 }
 
 static void seat_set_outputs_dirty(struct slurp_seat *seat) {
 	struct slurp_output *output;
 	wl_list_for_each(output, &seat->state->outputs, link) {
-		if (box_intersect(&output->logical_geometry, &seat->selection)) {
+		if (box_intersect(&output->logical_geometry,
+			&seat->pointer_selection.selection) ||
+		    box_intersect(&output->logical_geometry,
+			&seat->touch_selection.selection)) {
 			set_output_dirty(output);
 		}
 	}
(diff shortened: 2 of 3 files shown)


include/slurp.h
```diff
@@ -81,11 +89,13 @@ struct slurp_seat {
...
 
 	//selection (pointer/touch):
-	struct slurp_output *current_output;
+  /*struct slurp_output *current_output;
 	int32_t x, y;
 	int32_t anchor_x, anchor_y;
 	struct slurp_box selection;
-	bool has_selection;
+	bool has_selection;*/
+  struct slurp_selection pointer_selection;
```
emersion, line 97: Style: we use tabs, not spaces

include/slurp.h
```diff
@@ -81,11 +89,13 @@ struct slurp_seat {
 	struct wl_keyboard *wl_keyboard;
 
 	//selection (pointer/touch):
```
emersion, line 91: Can we drop these comments?

include/slurp.h
```diff
@@ -101,4 +111,5 @@ struct slurp_seat {
 };
 
 bool box_intersect(const struct slurp_box *a, const struct slurp_box *b);
+struct slurp_selection* seat_get_current_selection(struct slurp_seat* seat);
```
emersion, line 114: Style: `*` on the right

main.c
```diff
@@ -26,21 +26,37 @@ bool box_intersect(const struct slurp_box *a, const struct slurp_box *b) {
...
 
 static void move_seat(struct slurp_seat *seat, wl_fixed_t surface_x,
 		wl_fixed_t surface_y) {
-	int x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
-	int y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+	struct slurp_selection *current_selection =
+		seat_get_current_selection(seat);
+	if (current_selection == NULL) {
+	  //return;
+	  current_selection = &seat->pointer_selection; //still need to track the mouse coordinates;
```
emersion, line 48: Style: we use tabs, not spaces

render.c
```diff
@@ -6,15 +6,16 @@
...
 #include "slurp.h"
 
-static void set_source_u32(cairo_t *cairo, uint32_t color) {
-	cairo_set_source_rgba(cairo,
-		(color >> (3*8) & 0xFF) / 255.0,
-		(color >> (2*8) & 0xFF) / 255.0,
-		(color >> (1*8) & 0xFF) / 255.0,
-		(color >> (0*8) & 0xFF) / 255.0);
+static void set_source_u32(cairo_t *cairo, uint32_t color)
+{
```
emersion, line 10: Style: `{` on the same line as the function prototype

render.c
```diff
@@ -6,15 +6,16 @@
...
+	cairo_set_source_rgba(cairo, (color >> (3 * 8) & 0xFF) / 255.0,
+			      (color >> (2 * 8) & 0xFF) / 255.0,
```
emersion, line 12: Style: we don't align, we indent continuation lines with one extra tab instead

render.c
```diff
@@ -26,39 +27,49 @@ void render(struct slurp_output *output) {
 	cairo_paint(cairo);
 
 	struct slurp_seat *seat;
-	wl_list_for_each(seat, &state->seats, link) {
-		if (!seat->wl_pointer) continue;
-		if (!seat->has_selection) {
+	wl_list_for_each(seat, &state->seats, link)
+	{
```
emersion, line 31: Style: `{` on the same line as the previous one

render.c
```diff
@@ -6,15 +6,16 @@
...
+{
+	cairo_set_source_rgba(cairo, (color >> (3 * 8) & 0xFF) / 255.0,
+			      (color >> (2 * 8) & 0xFF) / 255.0,
+			      (color >> (1 * 8) & 0xFF) / 255.0,
+			      (color >> (0 * 8) & 0xFF) / 255.0);
 }
 
-void render(struct slurp_output *output) {
+void render(struct slurp_output *output)
+{
```
emersion, line 18: Style: `{` on the same line as the previous one

main.c
```diff
@@ -70,13 +86,16 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
...
-	seat->current_output = NULL;
+	seat->pointer_selection.current_output = NULL;
 }
 
 static void seat_set_outputs_dirty(struct slurp_seat *seat) {
 	struct slurp_output *output;
 	wl_list_for_each(output, &seat->state->outputs, link) {
-		if (box_intersect(&output->logical_geometry, &seat->selection)) {
+		if (box_intersect(&output->logical_geometry,
+				  &seat->pointer_selection.selection) ||
```
emersion, line 96: Style: we don't align, we indent with a tab instead

main.c
emersion, line 48: Hmm, so if there's no current selection, why do we need to update the pointer selection?

main.c
```diff
@@ -26,21 +26,37 @@ bool box_intersect(const struct slurp_box *a, const struct slurp_box *b) {
...
+		return &seat->touch_selection;
+	} else {
+		return NULL;
+	}
+}
+
 static struct slurp_output *output_from_surface(struct slurp_state *state,
 	struct wl_surface *surface);
 
 static void move_seat(struct slurp_seat *seat, wl_fixed_t surface_x,
```
emersion, line 42: This function should probably take a `struct slurp_selection *`. When called from the pointer event handler, it'd be called with `pointer_selection`. When called from the touch event handler, it'd be called with `touch_selection`.

main.c
```diff
@@ -102,40 +121,51 @@ static int max(int a, int b) {
...
+		seat->pointer_selection.has_selection = false;
 
 		// find smallest box intersecting the cursor
 		struct slurp_box *box;
-		wl_list_for_each(box, &seat->state->boxes, link) {
-			if (in_box(box, seat->x, seat->y)) {
-				if (seat->has_selection &&
-						box_size(&seat->selection) < box_size(box)) {
+		wl_list_for_each(box, &seat->state->boxes, link)
+		{
```
emersion, line 158: Style: `{` on the same line as the previous one

main.c
```diff
@@ -177,11 +212,15 @@ static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
 		uint32_t serial, uint32_t time, uint32_t button,
 		uint32_t button_state) {
 	struct slurp_seat *seat = data;
+	if (seat->touch_selection.has_selection) {
+	  return;
```
emersion, line 216: Style: we use tabs, not spaces

main.c
```diff
@@ -177,11 +212,15 @@ static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
...
+	}
 
 	seat->button_state = button_state;
 
 	switch (button_state) {
 	case WL_POINTER_BUTTON_STATE_PRESSED:
+		seat->pointer_selection.has_selection = true;
```
emersion, line 223: This should be taken care of by `handle_selection_start`. `handle_selection_start` should take a `struct slurp_selection *` as an argument.

main.c
```diff
@@ -234,13 +273,15 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
...
+			seat->pointer_selection.has_selection = false;
+			seat->touch_selection.has_selection = false;
 			state->edit_anchor = false;
 			state->running = false;
 			break;
 
 		case XKB_KEY_space:
-			if (!seat->has_selection) {
+			if (!seat->pointer_selection.has_selection &&
+			    !seat->touch_selection.has_selection) {
```
emersion, line 284: Style: we don't alignm, we indent with two tabs instead (to make a clear difference between the if condition and block).

render.c
```diff
@@ -26,39 +27,49 @@ void render(struct slurp_output *output) {
...
+		if (current_selection == NULL)
+			continue;
+		if (!seat->wl_pointer)
+			continue;
+		if (!current_selection->has_selection) {
 			continue;
 		}
 
-		if(!box_intersect(&output->logical_geometry, &seat->selection)) {
+		if (!box_intersect(&output->logical_geometry,
```
emersion, line 42: Style: all lines below this one should be unchanged.

render.c
```diff
@@ -26,39 +27,49 @@ void render(struct slurp_output *output) {
...
 
 	struct slurp_seat *seat;
-	wl_list_for_each(seat, &state->seats, link) {
-		if (!seat->wl_pointer) continue;
-		if (!seat->has_selection) {
+	wl_list_for_each(seat, &state->seats, link)
+	{
+		struct slurp_selection *current_selection =
+			seat_get_current_selection(seat);
+		if (current_selection == NULL)
```
emersion, line 34: Style: braces are mandatory (I see the previous code had a mistake).

render.c
```diff
@@ -26,39 +27,49 @@ void render(struct slurp_output *output) {
...
+			continue;
+		if (!seat->wl_pointer)
+			continue;
+		if (!current_selection->has_selection) {
```
emersion, line 38: This should already be enforced by `seat_get_current_selection`, right?

render.c
```diff
@@ -26,39 +27,49 @@ void render(struct slurp_output *output) {
...
-	wl_list_for_each(seat, &state->seats, link) {
-		if (!seat->wl_pointer) continue;
-		if (!seat->has_selection) {
+	wl_list_for_each(seat, &state->seats, link)
+	{
+		struct slurp_selection *current_selection =
+			seat_get_current_selection(seat);
+		if (current_selection == NULL)
+			continue;
+		if (!seat->wl_pointer)
```
emersion, line 36: If the seat has a touch device but no pointer device, we should still draw the selection. So I think we should drop this check.

main.c
gg-rewrite: move_seat updates mouse coordinates which are the start of the selection when clicked. Otherwise the mouse selection will always start at 0,0, i.e. in the corner.

I know it's crutchy but right now I can't think of any other way.

main.c
gg-rewrite: It's also called from move_seat and render, besides its whole point is to determine which selection is active right now.

render.c
gg-rewrite: I'm sorry, I don't  understand this one.

render.c
gg-rewrite: I think by this and previous comment I'll rewrite or even remove the function entirely.

main.c
gg-rewrite: b3be2ebf17a959e59ea449e643ee953bace8a583

On second thought I've just added the slurp_selection argument to the relevant functions (move_seat, handle_selection*) and made a simple check in render as it was the only place left. seat_get_current_selection is thus unneeded and all the crutches are gone.

render.c
emersion: Looking at the diff below, there are whitespace changes unrelated to this PR. It previously broke style, it doesn't anymore.

render.c
```diff
@@ -27,38 +26,46 @@ void render(struct slurp_output *output) {
 
 	struct slurp_seat *seat;
 	wl_list_for_each(seat, &state->seats, link) {
-		if (!seat->wl_pointer) continue;
-		if (!seat->has_selection) {
+		struct slurp_selection *current_selection =
+			(seat->touch_selection.has_selection) ?
+				(&seat->touch_selection) :
```
emersion, line 31: Style: unnecessary parentheses

main.c
```diff
@@ -30,17 +30,19 @@ static struct slurp_output *output_from_surface(struct slurp_state *state,
 	struct wl_surface *surface);
 
 static void move_seat(struct slurp_seat *seat, wl_fixed_t surface_x,
-		wl_fixed_t surface_y) {
-	int x = wl_fixed_to_int(surface_x) + seat->current_output->logical_geometry.x;
-	int y = wl_fixed_to_int(surface_y) + seat->current_output->logical_geometry.y;
+	wl_fixed_t surface_y,
+	struct slurp_selection *current_selection) {
```
emersion, line 34: Style: indent arguments with one extra tab to be able to tell them apart from the function body

main.c
emersion: Nice!

PR #60, opened 2020-08-16
Add output selection

vesim987:
Hi, I added new mode to slurp for output selection. I had to update xdg-output-unstable-v1 to version 2, and also `_POSIX_C_SOURCE` to 200809 for strdup.

Right now it is setting the output name for both, point and output selection, the only difference is that in output selection it is setting the output geometry as the selection box.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1,4 +1,4 @@
-#define _POSIX_C_SOURCE 2
+#define _POSIX_C_SOURCE 200809L
 
 #include <errno.h>
 #include <stdio.h>
@@ -169,12 +169,26 @@ static void handle_selection_start(struct slurp_seat *seat,
 	struct slurp_state *state = seat->state;
 	current_selection->has_selection = true;
 	
-	if (state->single_point) {
-		state->result.x = current_selection->x;
-		state->result.y = current_selection->y;
-		state->result.width = state->result.height = 1;
+	if (state->single_point || state->output_selection) {
+        
+        struct slurp_output *output;
+        wl_list_for_each(output, &state->outputs, link) {
+            if(in_box(&output->logical_geometry, current_selection->x, current_selection->y)) {
+                break;
+            }
+        }
+        state->result.output = output;
+        
+        if(state->output_selection) {
+            state->result.selection = output->logical_geometry;
+        } else {
+            state->result.selection.x = current_selection->x;
+            state->result.selection.y = current_selection->y;
+            state->result.selection.width = state->result.selection.height = 1;
+        }
+
 		state->running = false;
-	} else {
+    } else {
 		current_selection->anchor_x = current_selection->x;
 		current_selection->anchor_y = current_selection->y;
 	}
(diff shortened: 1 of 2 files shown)


main.c
```diff
@@ -613,7 +635,7 @@ static void handle_global(void *data, struct wl_registry *registry,
 		create_output(state, wl_output);
 	} else if (strcmp(interface, zxdg_output_manager_v1_interface.name) == 0) {
 		state->xdg_output_manager = wl_registry_bind(registry, name,
-			&zxdg_output_manager_v1_interface, 1);
+			&zxdg_output_manager_v1_interface, 2);
```
tmccombs, line 638: This may break if the compositor only supports version 1. If that is a concern, I think `MIN(2, version)` would be better.

main.c
```diff
@@ -708,7 +738,7 @@ int main(int argc, char *argv[]) {
 
 	int opt;
 	char *format = "%x,%y %wx%h";
```
tmccombs, line 740: If the `-o` option is used, should a different default format be used that includes the output name (or maybe is only the output name)?

main.c
```diff
@@ -169,12 +169,26 @@ static void handle_selection_start(struct slurp_seat *seat,
...
-		state->result.width = state->result.height = 1;
+	if (state->single_point || state->output_selection) {
+        
+        struct slurp_output *output;
+        wl_list_for_each(output, &state->outputs, link) {
+            if(in_box(&output->logical_geometry, current_selection->x, current_selection->y)) {
+                break;
+            }
+        }
+        state->result.output = output;
```
tmccombs, line 180: Would it make sense to set this for a rectangular selection as well? You do run into the problem of what to do if there are multiple outputs, but the simplest think would be just to pick a point, like the top left, or center of the rectangle, and use that to pick the "best" output.

danshick: Slurp can take a predefined set of rectangles as input and limit possible selections to those rectangles (you can see this if you use the output selection example in the README).

I think the correct approach is to implement it this way, where there's no option to select anything other than the predefined rectangles which occupy each output entirely, and chosen by the rectangle the mouse is inside of.

danshick: Looks like you can just use `add_choice_box` like is done here https://github.com/emersion/slurp/blob/master/main.c#L761

tmccombs: > limit possible selections to those rectangles

That isn't quite right. At least from me trying it out in slurp 1.2.0, it seems it just creates predefined boxes you can select with a click, but it doesn't stop you from dragging your own box.

danshick: Hmmm, that sounds like a bug in the intended behavior, but @emersion would have to weigh in.

emersion: No, this is intentional: the input boxes are just hints, the user can ignore the hint. I'd be fine with adding an option to prevent that.

emersion:
As @tmccombs said, I'd prefer this to be done with the format string. We could add `%o`, which prints the name of the output the selected box is in.

main.c
danshick: Glad I asked.

main.c
vesim987: I think the `-o` feature w/o outputs names is kinda useless. But that is a question to @emersion.

emersion: I think it's fine to bump the minimum required version to 2.

emersion:
Superseded by #64

PR #63, opened 2020-08-23
Force select

tmccombs:
This PR contains three related changes:

1. Fix a bug where if you hover over a choice box, then move away the choice box stays highlighted
2. Always render choice boxes as a different color, to make it easier to identify them, and add an option to specify the color (this is what I'm least sure about).
3. Add an option to force the user to select a choicebox (disallow custom rectangles). This will interact with #60 and whichever one is merged last will probably need to be updated to at least add documentation on how they interact.

If you'd prefer, I can split these out into seperate PRs.

### DIFF
--- a/main.c
+++ b/main.c
@@ -13,6 +13,10 @@
 #include "slurp.h"
 #include "render.h"
 
+#define BG_COLOR 0xFFFFFF40
+#define BORDER_COLOR 0x000000FF
+#define SELECTION_COLOR 0x00000000
+
 static void noop() {
 	// This space intentionally left blank
 }
@@ -637,6 +641,7 @@ static const char usage[] =
 	"  -b #rrggbbaa Set background color.\n"
 	"  -c #rrggbbaa Set border color.\n"
 	"  -s #rrggbbaa Set selection color.\n"
+	"  -B #rrggbbaa Set option box color.\n"
 	"  -w n         Set border weight.\n"
 	"  -f s         Set output format.\n"
 	"  -o           Select a display output.\n"
@@ -736,9 +741,10 @@ int main(int argc, char *argv[]) {
 
 	struct slurp_state state = {
 		.colors = {
-			.background = 0xFFFFFF40,
-			.border = 0x000000FF,
-			.selection = 0x00000000,
+			.background = BG_COLOR,
+			.border = BORDER_COLOR,
+			.selection = SELECTION_COLOR,
+			.choice = BG_COLOR,
 		},
 		.border_weight = 2,
 		.display_dimensions = false,
@@ -747,7 +753,7 @@ int main(int argc, char *argv[]) {
 	int opt;
 	char *format = "%x,%y %wx%h\n";
 	bool output_boxes = false;
-	while ((opt = getopt(argc, argv, "hdb:c:s:w:pof:")) != -1) {
+	while ((opt = getopt(argc, argv, "hdb:c:s:B:w:pof:")) != -1) {
 		switch (opt) {
 		case 'h':
 			printf("%s", usage);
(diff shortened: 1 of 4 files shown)


render.c
```diff
@@ -34,7 +34,11 @@ void render(struct slurp_output *output) {
 		if (!seat->wl_pointer) {
```
emersion: Unrelated to this PR, but: do we still need this check? It sounds like a touch-only device wouldn't render any selection because of this.

slurp.1.scd
```diff
@@ -40,6 +40,10 @@ held, the selection is moved instead of being resized.
 *-s* _color_
 	Set selection color. See *COLORS* for more detail.
 
+*-o* _color_
```
emersion, line 43: Why `-o`?

render.c
```diff
@@ -13,6 +13,17 @@ static void set_source_u32(cairo_t *cairo, uint32_t color) {
 		(color >> (0 * 8) & 0xFF) / 255.0);
 }
 
+static void draw_rect(cairo_t *cairo, struct slurp_box *box, uint32_t color, int32_t scale) {
+	set_source_u32(cairo, color);
+	cairo_rectangle(cairo, box->x * scale, box->y * scale,
+			box->width * scale, box->height * scale);
+}
+
+static void adjust_box(struct slurp_box *box, struct slurp_output *output) {
```
emersion, line 22: Maybe `box_layout_to_output` would be a better name?

main.c
```diff
@@ -706,14 +713,16 @@ int main(int argc, char *argv[]) {
 			.background = 0xFFFFFF40,
 			.border = 0x000000FF,
 			.selection = 0x00000000,
+			.choice = 0xCCCCFF40,
```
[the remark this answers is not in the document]
emersion: >Maybe it should be the same as the background color, and users can opt-in to always highlight the choice boxes by specifying a different color?

I'd prefer to use a fully transparent color, but then it wouldn't paint the background color underneath, correct?

In this case, yeah, I think default = background would be fine.

>Or maybe in practice the choice boxes usually always cover the entire screen, and this feature isn't useful at all?

That's not the case in general, choice boxes can be used to select e.g. windows (see README). Sometimes boxes are adjacent to the output edges though.

slurp.1.scd
```diff
@@ -14,8 +14,8 @@ slurp is a command-line utility to select a region from Wayland compositors
 which support the layer-shell protocol. It lets the user hold the pointer to
 select, or click to cancel the selection.
 
-If the standard input is not a TTY, slurp will read a list of predefined
-rectangles for quick selection. Each line must be in the form
+If the standard input is not a TTY or the -r option is used, slurp will read a
```
emersion, line 17: Unrelated to this PR: maybe we should have an option to force slurp to read from stdin, even if stdin is a TTY.

slurp.1.scd
tmccombs: whoops, this hould be `-B`.

render.c
tmccombs: Probably not. Unfortunately, I don't have a touch device that I can test this on.

slurp.1.scd
tmccombs: That's probably a good idea. Maybe I'll make a seperate pr for that.

emersion:
This needs a rebase.

tmccombs:
> This needs a rebase.

done.

render.c
emersion: Opened #66 as a follow-up issue.

slurp.1.scd
emersion: Opened #67 for this.

emersion (approved):
LGTM, thanks! Sorry for the long delay.

PR #64, opened 2020-08-24
Output format

tmccombs:
An alternative implementation for #60. Based on feedback on that PR.

### DIFF
--- a/main.c
+++ b/main.c
@@ -443,6 +443,7 @@ static void xdg_output_handle_logical_position(void *data,
 	output->logical_geometry.x = x;
 	output->logical_geometry.y = y;
 }
+
 static void xdg_output_handle_logical_size(void *data,
 		struct zxdg_output_v1 *xdg_output, int32_t width, int32_t height) {
 	struct slurp_output *output = data;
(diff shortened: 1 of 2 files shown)


main.c
```diff
@@ -873,6 +905,16 @@ int main(int argc, char *argv[]) {
 	// second roundtrip for xdg-output
 	wl_display_roundtrip(state.display);
 
+	if (output_boxes) {
+		struct slurp_output *box_output;
+		wl_list_for_each(box_output, &state.outputs, link) {
+			add_choice_box(&state, &box_output->logical_geometry);
+			struct slurp_box *new_box;
+			new_box = wl_container_of(state.boxes.prev, new_box, link);
+			new_box->label = strdup(new_box->label); // to avoid double free
```
tmccombs, line 914: I'm not sure if it's better to do this, or intentionally leak the labels. I think leaking the labels would probably be ok, since at that point we've reached the end of the program anyway.

main.c
```diff
@@ -490,6 +494,9 @@ static void destroy_output(struct slurp_output *output) {
 		wl_callback_destroy(output->frame_callback);
 	}
 	wl_output_destroy(output->wl_output);
+	if (output->logical_geometry.label) {
```
emersion, line 497: No need for this NULL check -- `free(NULL)` is a no-op.

tmccombs: oh, nice.

emersion:
I rebased the branch against master and squashed the last commit.

The first commit now LGTM.

main.c
emersion: I think `add_choice_box` should perform the `strdup` instead.

emersion (approved):
I squashed the last commit.

This now LGTM, thanks!

tmccombs:
Thanks for making such a useful tool.

PR #65, opened 2020-08-25
Allow a custom formatting to not have a newline in the end

SeppeSoete:
This makes output it a bit more easily readable in scripting.

### DIFF
--- a/main.c
+++ b/main.c
@@ -686,7 +686,7 @@ static void print_formatted_result(const struct slurp_box *result,
 		}
 		printf("%c", c);
 	}
-	printf("\n");
+    fflush(stdout);
 }
 
 static void add_choice_box(struct slurp_state *state,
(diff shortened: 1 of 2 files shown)


slurp.1.scd
```diff
@@ -68,7 +68,7 @@ Interpreted sequences are:
 
 %l	Label included with region from stdin
 
-The default format is "%x,%y %wx%h".
+The default format is "%x,%y %wx%h\n".
```
emersion, line 71: This renders as `%x,%y %wx%hn`. We need to escape the `\`.

main.c
```diff
@@ -686,7 +686,6 @@ static void print_formatted_result(const struct slurp_box *result,
 		}
 		printf("%c", c);
 	}
-	printf("\n");
```
emersion, line 689: Would maybe be a good idea to `fflush(stdout)` here

slurp.1.scd
SeppeSoete: resolved in 9966d67

main.c
SeppeSoete: resolved in 630e1ad

emersion (approved):
Thanks, this LGTM! Can you squash all of these commits into a single one?

SeppeSoete:
Squashed into a single commit as requested.

emersion (approved):
Thanks!

PR #69, opened 2020-10-17
Print result after destroying output.

grmat:
Hi,

after updating to 1.3.0, I've noticed that my screenshots now have black borders around the edges. I bisected this behavioural change to 1c843566c53544b18648adee73d401ca9e1ae040 which moved the call of print_formatted_result(). If the output is piped into another tool to capture the screen, this results in the drawn border being part of the capture which is likely unwanted.

### DIFF
--- a/main.c
+++ b/main.c
@@ -671,23 +671,23 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
-static void print_output_name(const struct slurp_box *result, struct wl_list *outputs) {
+static void print_output_name(FILE *stream, const struct slurp_box *result, struct wl_list *outputs) {
 	struct slurp_output *output;
 	wl_list_for_each(output, outputs, link) {
 		// For now just use the top-left corner
 		struct slurp_box *geometry = &output->logical_geometry;
 		if (in_box(geometry, result->x, result->y)) {
 			if (geometry->label) {
-				printf("%s", geometry->label);
+				fprintf(stream, "%s", geometry->label);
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -990,5 +982,12 @@ int main(int argc, char *argv[]) {
 		free(box);
 	}
 
+	if (state.result.width == 0 && state.result.height == 0) {
+		fprintf(stderr, "selection cancelled\n");
+		status = EXIT_FAILURE;
+	} else {
+		print_formatted_result(&state.result, &state.outputs, format);
```
emersion, line 989: This won't do, `print_formatted_result` needs access to `state.outputs`.

grmat: Of course you're right. Excuse my naive approach.
Would it be acceptable approach for `print_formatted_result` (maybe rename it) to return the string instead of printing it, then printing it from the main function after destroying the outputs?

emersion: >Of course you're right. Excuse my naive approach.

No worries!

>Would it be an acceptable approach for print_formatted_result (maybe rename it) to return the string instead of printing it, then printing it from the main function after destroying the outputs?

Yeah, this sounds reasonable. The tricky part is that we need a big enough buffer. I'd suggest looking into `open_memstream` so that libc handles all of the dynamic buffer management and we can continue to use `fprintf`.

emersion (approved):
Thanks!

grmat:
Thank you!

PR #74, opened 2020-11-04
Fix off-by-one error in in_box function

David96:
Because the coordinate is 0 indexed, "greater than", not "greater or
equal than" has to be used here (imagine a 1x1 box, it only contains the
coordinates (0,0), not (1,1)).

This fixes output selection printing the wrong display name (as it uses
the upper left corner of the selection to figure out the display, it
falls into this edge case).


emersion (approved):
LGTM, and the Git history doesn't seem to indicate that the checks were `>=` for a real reason.

Thanks!

PR #77, opened 2020-11-08
Add option to enforce a specific aspect ratio of the selection

### LINKED ISSUE #75: Feature: Fixed aspect ratio?
I just started playing around with a graphics tablet on sway, and figured that it might be nice to use slurp to set mappings for tablet input, e.g.

```
# sway.conf

bindsym $mod+t exec swaymsg "input type:tablet_tool map_to_region $(slurp -f \"%x %y %w %h\")"
```

For this usecase it would however be useful to fix the aspect ratio of the selection to that of the tablet. Thus, I'd like to propose adding an `--aspect <ratio>` option that implements this behaviour. Could this be added? If yes, I might look into implementing it myself. I've essentially no knowledge of sway/slurp/wayland internals, though, so if someone else would go ahead, I'd appreciate it a lot. Thanks!

wisp3rwind:
Implements the feature proposed in #75. Thanks for considering to add this to slurp! It would be good if someone could verify that I'm not adding a new off-by-one error here.

Closes #75

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -55,6 +55,7 @@ struct slurp_state {
 	bool single_point;
 	bool restrict_selection;
 	struct wl_list boxes; // slurp_box::link
+	bool fixed_aspect_ratio;
 	double aspect_ratio;  // h / w
 
 	struct slurp_box result;
@@ -113,4 +114,10 @@ struct slurp_seat {
 };
 
 bool box_intersect(const struct slurp_box *a, const struct slurp_box *b);
+
+static inline struct slurp_selection *slurp_seat_current_selection(struct slurp_seat *seat) {
+	return seat->touch_selection.has_selection ?
+		&seat->touch_selection :
+		&seat->pointer_selection;
+}
 #endif
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -53,6 +53,8 @@ struct slurp_state {
 	bool single_point;
 	bool restrict_selection;
 	struct wl_list boxes; // slurp_box::link
+	bool fixed_aspect;
+	double aspect_ratio;  // h / w
 
 	struct slurp_box result;
 };
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -55,7 +55,6 @@ struct slurp_state {
 	bool single_point;
 	bool restrict_selection;
 	struct wl_list boxes; // slurp_box::link
-	bool fixed_aspect;
 	double aspect_ratio;  // h / w
 
 	struct slurp_box result;
--- a/main.c
+++ b/main.c
@@ -132,16 +132,25 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 	seat->pointer_selection.current_output = NULL;
 }
 
-static void handle_active_selection_motion(struct slurp_seat *seat, struct slurp_selection *current_selection) {
+static void handle_active_selection_motion(struct slurp_seat *seat,
+		struct slurp_selection *current_selection) {
 	int32_t anchor_x = max(0, current_selection->anchor_x);
 	int32_t anchor_y = max(0, current_selection->anchor_y);
-	current_selection->selection.x = min(anchor_x, current_selection->x);
-	current_selection->selection.y = min(anchor_y, current_selection->y);
+	int32_t dist_x = current_selection->x - anchor_x;
+	int32_t dist_y = current_selection->y - anchor_y;
+
 	// selection includes the seat and anchor positions
-	current_selection->selection.width =
-		abs(current_selection->x - anchor_x) + 1;
-	current_selection->selection.height =
-		abs(current_selection->y - anchor_y) + 1;
+	int32_t width = abs(dist_x) + 1;
+	int32_t height = abs(dist_y) + 1;
+	if (seat->state->fixed_aspect) {
+		width = min(width, height / seat->state->aspect_ratio);
+		height = min(height, width * seat->state->aspect_ratio);
+	}
+
+	current_selection->selection.x = dist_x > 0 ? anchor_x : anchor_x - (width - 1);
+	current_selection->selection.y = dist_y > 0 ? anchor_y : anchor_y - (height - 1);
+	current_selection->selection.width = width;
+	current_selection->selection.height = height;
 }
 
 static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
@@ -651,7 +660,8 @@ static const char usage[] =
 	"  -f s         Set output format.\n"
 	"  -o           Select a display output.\n"
 	"  -p           Select a single point.\n"
-	"  -r           Restrict selection to predefined boxes.\n";
+	"  -r           Restrict selection to predefined boxes.\n"
+	"  -a w:h       Force aspect ratio.\n";
 
 uint32_t parse_color(const char *color) {
 	if (color[0] == '#') {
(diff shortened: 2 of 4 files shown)


LeonardDrs:
Hey nice!
Somehow related question: What about the "usual" feature of forcing the aspect ratio to 1:1 by holding `Shift`?

wisp3rwind:
> Somehow related question: What about the "usual" feature of forcing the aspect ratio to 1:1 by holding `Shift`?

Do you really mean 1:1, or that tablet and screen area have the same aspect ratio? In what sense is holding Shift "usual"? In any case, that's probably not too hard to implement. I'm not sure how much feature creep the slurp maintainers are comfortable with, though.

> Second somehow related question: Does the `hold space to move` still keeps the ratio as intended?

It should. I think slurp crops the selection when you reach the top/left screen edges, in that case the aspect ratio might be incorrect until you release space and move the pointer again, I'd have to check.

LeonardDrs:
> Do you really mean 1:1, or that tablet and screen area have the same aspect ratio? In what sense is holding Shift "usual"? In any case, that's probably not too hard to implement. I'm not sure how much feature creep the slurp maintainers are comfortable with, though.

I meant really 1:1, like when you're trying to do a selection on drawing software, you can often hold a modifier key to ensure your selection is going to be a square.
There is also sometimes another modifier key to select from the center instead of the top-left, but I don't yet feel the need for such selection behavior.

Pict: 
recording

Anyway I feel like I'm hijacking your PR for this, sorry!

luissantos:
This feature is great. I can confirm that it works as expected. 
Any plans to get it merged? 
Thanks

tmccombs:
Unfortunately it looks like there are some conflicts on it now.

include/slurp.h
```diff
@@ -53,6 +53,8 @@ struct slurp_state {
 	bool single_point;
 	bool restrict_selection;
 	struct wl_list boxes; // slurp_box::link
+	bool fixed_aspect;
```
tmccombs, line 56: we could probably use an `aspect_ratio` of 0 to indicate no fixed aspect ration, since it doesn't really make sense to have an aspect ratio of 0

wisp3rwind:
I'm not using sway/slurp/a tablet at the moment, so I'll not be pushing this PR further. Feel free to take it over, and use any of the code.

include/slurp.h
layus: But then, it is also good not having to think too much about corner cases. And using a 0.0 on a double may be annoying at some point.

main.c
```diff
@@ -104,14 +104,21 @@ static void handle_active_selection_motion(struct slurp_seat *seat, struct slurp
 
 	int32_t anchor_x = current_selection->anchor_x;
 	int32_t anchor_y = current_selection->anchor_y;
+  int32_t dist_x = current_selection->x - anchor_x;
```
layus, line 107: There seems to be several buggy indents. Tabs vs spaces issue maybe ?

main.c
```diff
@@ -104,14 +104,21 @@ static void handle_active_selection_motion(struct slurp_seat *seat, struct slurp
...
 	// selection includes the seat and anchor positions
-	current_selection->selection.width =
-		abs(current_selection->x - anchor_x) + 1;
-	current_selection->selection.height =
-		abs(current_selection->y - anchor_y) + 1;
+  int32_t width = abs(dist_x) + 1;
+	int32_t height = abs(dist_y) + 1;
+ 	if (seat->state->fixed_aspect) {
+		width = min(width, height / seat->state->aspect_ratio);
+		height = min(height, width * seat->state->aspect_ratio);
```
layus, line 116: I think we want `max` here, as we want the cursor to be always inside the selection, not outside.
See for example fixed ratio selection in gimp with Shift key.

selection

layus (approved):
Thanks for doing the job, and for rebasing it. I think there is decision to take about min/max selection, but otherwise this is ready to merge.

main.c
layus: Well, at least min does not allow to select outside screen borders. It may be a reasonable property here.

tmccombs:
I've added bounds checking switched it to using max instead of min, and added support for pressing shift to get a 1:1 aspect ratio.

layus:
Thanks, I have tested it and it is a really nice improvement on the ui. And thanks again for the good work. :tada: 

I am still uneasy about the Shift-bound 1:1 ratio, but do not let that stop shipping this feature.
I wonder when such a hard-coded ratio may come in handy, and why 1:1 over more common pictures ratios, or the ratio of the current monitor, or the full monitor grid.  And I fail to see the right usage patter, or any rationale.

Coming to think of it, Shift should probably _disable_ the current ratio constraint, so that users can change their mind and get rid of the ratio if it does not work with what they are selecting. Then `-g 1:1` can be added by default by users so inclined, and remains escapeable with Shift if needed.

LeonardDrs:
>And I fail to see the right usage patter, or any rationale.

Never wanted to select a perfect square? Because every time I do, I spend 5-10s having the perfect ratio.

>Then -g 1:1 can be added by default by users so inclined

I feel like having a cli option for this make the feature harder to use. What if Gimp required you to go to the settings to select with a ratio?
Also I don't feel like having a constraint by default but escapeable are common. I'd rather have an opt-in over an opt-out.
That's only my opinion.

layus:
> > And I fail to see the right usage patter, or any rationale.
> 
> Never wanted to select a perfect square? Because every time I do, I spend 5-10s having the perfect ratio.

Oh, yes, sometimes I do, but sometimes I want another ratio. It just does not feel right to get that as the arbitrary default ratio. And when you want a perfect square, you know it in advance, so it fits well with the new `-g` option.
 
> > Then -g 1:1 can be added by default by users so inclined
> 
> I feel like having a cli option for this make the feature harder to use. What if Gimp required you to go to the settings to make a selection with a 1:1 ratio? Also I don't feel like having a constraint by default but escapeable is common for the selection behavior. I'd rather have an opt-in over an opt-out. That's only my opinion, which is biased over the usage of photoshop and osx selection feature though

Gimp does a 1:1 box selection on Shift, _by default_, but the crop tool uses the ratio of the current canvas by default. So it is not always 1:1. For Gimp it does not matter, as the ratio can be tuned in the UI. But here it feels like we are programming a feature and reserving an important key (Shift, no less) for an arbitrary ratio that is not certain to be more useful than others. We could also use shift to make the selection stick to 10px increments grid. I also sometimes want to have a nicely rounded number of pixels.

Also, it is a bit ironic that Shift allows to select a ratio that is different from the one in the command line, but that you can only get a square one. So Shift effectively acts as a partial escape from the `-g` command line argument, but only to fall into another jail ;-).

But again, do not let my questions delay in any way this PR. The feature is much needed, and fixes to Shift behavior can be discussed in another thread.

tmccombs:
> Also, it is a bit ironic that Shift allows to select a ratio that is different from the one in the command line, but that you can only get a square one. So Shift effectively acts as a partial escape from the -g command line argument, but only to fall into another jail ;-).

Actually, shift doesn't do anything if you specify the aspect ratio with `-a`.  At least with my current implementation. 

Personally, I think a 1:1 ratio would be the most common for me to want, and since that is what the rectangle tool in gimp does I went with that. But I suppose I don't have any reason to think that would be true for everyone.  But I'm not sure what a better default would be.

tmccombs:
I've added some documentation of keyboard controls to the man page, including the new shift behavior, but marked it as experimental to give us the flexibility to change it later if a better behavior is landed on. Any objection to merging with that?

tmccombs (approved):

layus (approved):
:tada: Ship it :-)

PR #79, opened 2021-02-13
Fix loss of selection when pointer moves between outputs

### LINKED ISSUE #72: lost selection after moving to another screen
Slurp(v1.3) lost its selection when a pointer is moved to other monitors.
I checked the version 1.2 has no problem.

bmwalters:
Thanks for this project!

This pull request contains the following fixes:
- Fixes #72
- Fixes incorrect anchor (locked to x = 0) when output has negative position 

See commit messages for details.

### DIFF
--- a/main.c
+++ b/main.c
@@ -96,6 +96,18 @@ static void seat_set_outputs_dirty(struct slurp_seat *seat) {
 	}
 }
 
+static void handle_active_selection_motion(struct slurp_seat *seat, struct slurp_selection *current_selection) {
+	int32_t anchor_x = current_selection->anchor_x;
+	int32_t anchor_y = current_selection->anchor_y;
+	current_selection->selection.x = min(anchor_x, current_selection->x);
+	current_selection->selection.y = min(anchor_y, current_selection->y);
+	// selection includes the seat and anchor positions
+	current_selection->selection.width =
+		abs(current_selection->x - anchor_x) + 1;
+	current_selection->selection.height =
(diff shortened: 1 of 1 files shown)


tmccombs (changes requested):
I've tested this a little, bit and most things work. However, removing setting `has_selection` to false at the beginning of `seat_update_selection` has the side effect that if there is a selectable region provided via stdin and you hover over it, it highlights the region, but when you leave the region, the highlight remains instead of going way. I think now we are just never unsetting `seat->pointer_selection.has_selection`.

bmwalters:
True. I changed `seat_update_selection` to be a no-op if there are no predefined rectangles.

tmccombs:
hmm, this has the interesting effect that if boxes are supplied on the command line, then selection is still lost if you drag a selection between outputs, but if no boxes are supplied then the selection isn't lost. 

I wonder if maybe there should be a distinction between a hover selection and a drag selection.

bmwalters:
> I wonder if maybe there should be a distinction between a hover selection and a drag selection.

It turns out that `pointer_handle_motion` makes this distinction by checking `seat->button_state`. I implemented this same logic in `pointer_handle_enter` and the patch seems to behave better now.

> hmm, this has the interesting effect that if boxes are supplied on the command line, then selection is still lost if you drag a selection between outputs, but if no boxes are supplied then the selection isn't lost.

Should be fixed in the latest revision.

emersion (approved):
Sorry for the late review. This looks good to me, thanks!

PR #88, opened 2021-06-03
has_selection is true only after moving the mouse

### LINKED ISSUE #86: Clicking outside all rectangles should cancel
When using `slurp` with sway with multiple monitors to screenshot a single window, the following feels unintuive to me:

- Monitor 1: one window open
- Monitor 2: empty workspace
- run `grimshot save window`, which passes the window rectangles to slurp, e.g: `echo -e "300,100 600x600\n900,100 600x600" | slurp`
- Hover the window on monitor 1
- Hover the empty workspace on monitor 2
- The window on monitor 1 is now grayed out
- Click somewhere on monitor 2
- Slurp has selected the window on monitor 1, instead of being cancelled.

Expected behavior: slurp cancels when clicking outside all rectangles
Actual behavior: slurp selects the last hovered rectangle when clicking outside all rectangles

The fix is probably to explicitly clear any pending selection when the mouse moves to an area not covered by any rectangles.

RagnarGrootKoerkamp:
Fix #86.

Old behavior:
When the mouse is pressed, `has_selection` is set to `true`, and the `anchor` is set, but the selection itself is not actually filled. When the mouse is release, `has_selection` is true so the [outdated] value of `selection` is read.

New behavior:
`has_selection` is only set to true when the `selection` is set, i.e. on mouse move.

All of `-p`, `-r` and `slurp` without arguments still work fine, but some testing may be needed - I don't fully understand all the logic myself.

### DIFF
--- a/main.c
+++ b/main.c
@@ -99,6 +99,7 @@ static void seat_set_outputs_dirty(struct slurp_seat *seat) {
 static void handle_active_selection_motion(struct slurp_seat *seat, struct slurp_selection *current_selection) {
 	int32_t anchor_x = current_selection->anchor_x;
 	int32_t anchor_y = current_selection->anchor_y;
+	current_selection->has_selection = true;
 	current_selection->selection.x = min(anchor_x, current_selection->x);
 	current_selection->selection.y = min(anchor_y, current_selection->y);
 	// selection includes the seat and anchor positions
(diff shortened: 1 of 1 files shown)


tmccombs (changes requested):
This mostly seems good. 

One weird thing is that if you run something like

```
echo 0,0 300x400 | slurp -r
```
then start dragging outside of the selection box, it will show a selection anchored at 0,0 to the mouse location. Then when you release the selection goes away, because you didn't select a valid region. Possibly that could be remedied by only setting `has_selection` to `true` here if `state->restrict_selection` is false?

RagnarGrootKoerkamp:
Ah, good find. Yeah your suggestion seems good.

It could be taken further to never call that function when -r is passed in the first place, but this is cleaner.

tmccombs (approved):

PR #89, opened 2021-06-03
Mark moved-away from outputs as dirty

### LINKED ISSUE #87: Moving the mouse to a different output does not always grey out rectangles
When using sway with two monitors, with one or more windows active on both monitors, the following bug occurs:

- Have one (or more) windows open on each monitor
- Run `grimshot save window`, which calls `slurp -r` and passes in the rectangles for all windows.
- Moving the mouse (quickly) back and forth between the two outputs occasionally (~50% of times) does not blur/gray out windows.

Expected behavior: When the mouse leaves a window, it is *always* grayed out
Actual behavior: When the mouse leaves a window (and moves to another output), the window is sometimes not grayed out.

Note 1: everything works fine when quickly moving the mouse between windows on the same output.
Note 2: This happens both with and without using gaps in sway. (I.e. it doesn't matter whether the two rectangles exactly touch or not.)

RagnarGrootKoerkamp:
Just like how pointer_handle_motion marks the old state as dirty,
pointer_handle_enter should also mark the old state as dirty.
This fixes #87.
Not sure whether leave or enter is preferred, but enter seems nicer
because that's closer to where the state is changed.

There is still a race condition left, although it's very unlikely to
happen:
1. handle_motion or handle_enter is called. The current selection is on output A.
2. The output A is marked dirty
3. render() is called async, and handles the redraw of A, using the old
   selection.
4. render() marks A a clean
5. move_seat is called from handle_motion or handle_enter. This updates
   the selection to point to output B and B is marked dirty
6. B will be redraw on further events, but A will forever remain in a

### DIFF
--- a/main.c
+++ b/main.c
@@ -116,6 +116,17 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 	if (output == NULL) {
 		return;
 	}
+
+	// the places the cursor moved away from are also dirty
+	// TODO: This should dirty outputs the current value
+	// pointer_selection.selection _after_ calling move_seat.  Currently there
+	// is a race condition when render executes after outputs are dirtied using
+	// the current value, but before the selection is updated. In that case, the
+	// the current output will never be dirtied for the new value of pointer_selection.
+	if (seat->pointer_selection.has_selection) {
+		seat_set_outputs_dirty(seat);
+	}
+
 	// TODO: handle multiple overlapping outputs
 	seat->pointer_selection.current_output = output;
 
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -116,6 +116,17 @@ static void pointer_handle_enter(void *data, struct wl_pointer *wl_pointer,
 	if (output == NULL) {
 		return;
 	}
+
+	// the places the cursor moved away from are also dirty
+	// TODO: This should dirty outputs the current value
```
tmccombs, line 121: isn't slurp single-threaded? is it even possible for `render()` to be called at that point?

tmccombs (approved):

main.c
RagnarGrootKoerkamp: Oh - good question. I had just assumed it's multithreaded since it's handling events and setting callbacks, but I have no idea how Wayland works really. I could just take these TODOs out and things will be fine in practice.

tmccombs: @emersion would know better than me, but my understanding is that it is single threaded, but asynchronous. So the render would happen on a subsequent iteration of the loop that calls `wl_display_dispatch`.

PR #94, opened 2021-09-11
Use cairo_scale instead of multiplying by the scale everywhere.

### LINKED ISSUE #93: Simplify HiDPI rendering code
Instead of having `* scale` all over the place, we can just use `cairo_scale` to set the scale factor.

Same as https://github.com/swaywm/sway/issues/6486

tmccombs:
Fixes #93

### DIFF
--- a/pool-buffer.c
+++ b/pool-buffer.c
@@ -64,7 +64,7 @@ static const struct wl_buffer_listener buffer_listener = {
 };
 
 static struct pool_buffer *create_buffer(struct wl_shm *shm,
-		struct pool_buffer *buf, int32_t width, int32_t height) {
+		struct pool_buffer *buf, int32_t width, int32_t height, double scale) {
 	const enum wl_shm_format wl_fmt = WL_SHM_FORMAT_ARGB8888;
 	const cairo_format_t cairo_fmt = CAIRO_FORMAT_ARGB32;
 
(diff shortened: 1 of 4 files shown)


pool-buffer.c
```diff
@@ -120,7 +122,11 @@ void finish_buffer(struct pool_buffer *buffer) {
 }
 
 struct pool_buffer *get_next_buffer(struct wl_shm *shm,
-		struct pool_buffer pool[static 2], uint32_t width, uint32_t height) {
+		struct pool_buffer pool[static 2], uint32_t width, uint32_t height, double scale) {
```
emersion, line 125: I'd rather not pass down the scale to `get_next_buffer`. `get_next_buffer` doesn't need to care about scale. Instead we can just reset the scale right before rendering:

```c
cairo_identity_matrix(cairo);
cairo_scale(cairo, scale, scale);
```

pool-buffer.c
```diff
@@ -132,12 +138,20 @@ struct pool_buffer *get_next_buffer(struct wl_shm *shm,
...
+	if (buffer->width != buffer_width || buffer->height != buffer_height) {
 		finish_buffer(buffer);
+	} else if (buffer->scale != scale) {
+		// This is a matrix that multiples both x and y values by scale.
+		cairo_matrix_t scale_matrix = {
+			scale, 0,
+			0,     scale,
+			0,     0,
+		};
+		cairo_set_matrix(buffer->cairo, &scale_matrix);
```
emersion, line 150: This can be simplified with `cairo_scale`.

emersion (approved):
LGTM, thanks!

PR #56, opened 2020-05-06
Add support for crosshairs

ammgws:
Resolves #45 

Draws fullscreen crosshairs on the output with the cursor until a selection is started.

PR #95, opened 2021-09-26
Crosshairs

tmccombs:
This is based on https://github.com/emersion/slurp/pull/56, but has been rebased on master, and fixed some issues I found while testing (mostly around outputs that don't start at the origin).


main.c
```diff
@@ -85,12 +67,14 @@ static void seat_update_selection(struct slurp_seat *seat) {
...
 static void seat_set_outputs_dirty(struct slurp_seat *seat) {
+	struct slurp_state *state = seat->state;
 	struct slurp_output *output;
 	wl_list_for_each(output, &seat->state->outputs, link) {
 		if (box_intersect(&output->logical_geometry,
 			&seat->pointer_selection.selection) ||
 		    box_intersect(&output->logical_geometry,
-			&seat->touch_selection.selection)) {
+			&seat->touch_selection.selection) ||
+			state->crosshairs) {
```
tmccombs, line 77: this should probably check that the selection is on the output, but I haven't gotten that to work yet.

tmccombs:
> When moving the crosshairs between monitors, sometimes this happens:

That's unfortunate. I'm not entirely sure why that happens. Reverting  0eeba3ae32eaecc2c36a6bfb5df363880c9adf91 fixes, it. That does mean we mark the outputs as dirty more often than necessary. I'd have to investigate more to see if there is a better way to handle that.

ammgws:
By the way do you have merge permissions or does this still require @emersion's approval first?

tmccombs:
I have merge permissions, but I would like it to be reviewed by someone else before merging.

ammgws:
>  Reverting 0eeba3a fixes, it. That does mean we mark the outputs as dirty more often than necessary. I'd have to investigate more to see if there is a better way to handle that.

Any luck with this? Or perhaps this could still be shipped since it doesn't seem like a showstopper, especially since crosshairs are not enabled by default.

tmccombs:
No, I haven't investigated it any more yet. (But the revert is on this branch).

tmccombs:
Ok, I think I've got better checks on when to mark outputs as dirty now. I was just missing a condition in an if statements one of the places where `seat_set_outputs_dirty` was called.

ammgws:
This has been working well for me, thanks!

tmccombs:
@emersion any objection to merging this?

ammgws:
@tmccombs any luck?

tmccombs:
I've resolved the conflicts.

alexmaras:
I've been using this locally for a while from a manual build and I've been super happy with it. Thanks for putting it together @tmccombs!

danielwerg:
Thank you, love this feature! I just have a small issue, afaik there is no way to change width/weight of the lines. `-w` doesn't apply unlike `-c` (this should be in man page btw). It's a bit too thin for my liking.

include/slurp.h
```diff
@@ -5,122 +5,117 @@
...
-	bool display_dimensions;
-	bool single_point;
-	bool restrict_selection;
-	bool resizing_selection;
-	struct wl_list boxes; // slurp_box::link
-	bool fixed_aspect_ratio;
-	double aspect_ratio;  // h / w
-
-	struct slurp_box result;
+  bool running;
```
emersion, line 25: Code style: this PR replaces all tabs by spaces. We use tabs, not spaces.

PR #97, opened 2021-11-06
Render dimensions with same color as the border

rjimenezda:
~This PR adds a `-D` argument that lets the user specify the text color for the displayed dimensions.~

This PR adds the ability to use the border color for the text rendering

https://user-images.githubusercontent.com/446970/140625682-0846d26f-584c-474d-903c-06e9964d4ff5.mp4


emersion:
Maybe we can just use `state.colors.border` instead?

rjimenezda:
@emersion Good point!

I just undid most of the changes and pushed a clean commit doing just that :smile:

emersion (approved):
Thanks!

PR #99, opened 2021-11-09
render: Allow the user to specify which font family to use

pedrolucasp:
User can pass the `-F` option and specify one of standard CSS2 generic
font family names. It'll default to the sans family if not specified and
only be visible in combination with the `-d` option.

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -48,6 +48,8 @@ struct slurp_state {
 		uint32_t choice;
 	} colors;
 
+	const char *font_family;
+
 	uint32_t border_weight;
 	bool display_dimensions;
 	bool single_point;
--- a/slurp.1.scd
+++ b/slurp.1.scd
@@ -44,6 +44,12 @@ held, the selection is moved instead of being resized.
 	Set color for highlighting predefined rectangles from standard input when not
 	selected.
 
+*-F* _font family_
+	Set the font family name when displaying the dimensions box. Only useful
+	when combined with the -d option. The available font family names guaranteed
+	to work are the standard generic CSS2 options: serif, sans-serif,
+	monospace, cursive and fantasy. It defaults to the sans-serif family name.
+
 *-w* _weight_
 	Set border weight.
 
--- a/main.c
+++ b/main.c
@@ -16,6 +16,7 @@
 #define BG_COLOR 0xFFFFFF40
 #define BORDER_COLOR 0x000000FF
 #define SELECTION_COLOR 0x00000000
+#define FONT_FAMILY "sans-serif"
 
 static void noop() {
 	// This space intentionally left blank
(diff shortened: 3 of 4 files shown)


slurp.1.scd
```diff
@@ -44,6 +44,12 @@ held, the selection is moved instead of being resized.
 	Set color for highlighting predefined rectangles from standard input when not
 	selected.
 
+*-F* _font family_
+	Set the font family name when displaying the dimensions box. Only useful
+	when combined with the -d option. The available font family names are the
+	standard generic CSS2 options: serif, monospace or sans.
```
tmccombs, line 50: The generic CSS2 options also include `fantasy` and `cursive`, and it is `sans-serif` not `sans` (see https://www.w3.org/TR/CSS2/fonts.html#generic-font-families). Other fonts families may also work, but there isn't any guarantee. I'm not sure if that is worth mentioning.

tmccombs (approved):
The implementation looks fine to me, and seems to work as expected. 

I am curious what the motivation for this is though.

include/slurp.h
```diff
@@ -48,6 +48,8 @@ struct slurp_state {
 		uint32_t choice;
 	} colors;
 
+	char *font_family;
```
emersion, line 51: This can be `const`, to make it clearer that it shouldn't be free'd.

slurp.1.scd
pedrolucasp: Right, I've updated my patch to mention about the other generic font family names and replaced to use `sans-serif` instead of `sans` (I just kept like that because it was being used on the original implementation).

include/slurp.h
pedrolucasp: Nice, updated my patch accordingly.

slurp.1.scd
```diff
@@ -44,6 +44,12 @@ held, the selection is moved instead of being resized.
 	Set color for highlighting predefined rectangles from standard input when not
 	selected.
 
+*-F* _font family_
+	Set the font family name when displaying the dimensions box. Only useful
+	when combined with the -d option. The available font family names are the
+	standard generic CSS2 options: serif, sans-serif, monospace, cursive and
+	fantasy. Defaults to the Sans-Serif Family name.
```
emersion, line 51: This makes it sound like "real" font names won't work, can we reword that?

pedrolucasp: Well, if anything, the docs do mention that this API has no guarantee of working with "real" fonts.

> Selects a family and style of font from a simplified description as a family name, slant and weight. Cairo provides no operation to list available family names on the system (this is a "toy", remember), but the standard CSS2 generic family names [...] are likely to work as expected. [...] For "real" font selection, see the font-backend-specific `font_face_create` functions for the font backend you are using.

Do you still want me to reword anyway?

emersion: Well, it just says that there's no way to get a list of the font names which will work, since that depends on the system's installed fonts. But "real" font names will definitely work fine. The CSS2 generic names just happen to be installed and configured on most systems.

IOW, if the user installs "Liberation", then configures slurp to use that font, it'll work fine.

emersion (approved):
Thanks!

PR #101, opened 2022-01-14
Add format string variables for output-local coordinates

puffnfresh:
For example, if you select from the top-left of the current display,
you will get a (0,0) coordinate, no matter where the output is
configured on the surface.

This is useful for the many tools which want coordinates relative to
the current output.

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -13,6 +13,7 @@
 
 struct slurp_box {
 	int32_t x, y;
+	int32_t output_local_x, output_local_y;
 	int32_t width, height;
 	char *label;
 	struct wl_list link;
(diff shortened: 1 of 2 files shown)


main.c
```diff
@@ -721,6 +727,12 @@ static void print_formatted_result(FILE *stream, const struct slurp_box *result,
 			case 'y':
 				fprintf(stream, "%d", result->y);
 				continue;
+			case 'X':
```
tmccombs, line 730: Could you add a description of these options to the man page (edit slurp.1.scd) in the FORMAT section?

tmccombs (changes requested):
Besides updating the man page, this looks good to me.

include/slurp.h
```diff
@@ -13,6 +13,7 @@
 
 struct slurp_box {
 	int32_t x, y;
+	int32_t output_local_x, output_local_y;
```
emersion, line 16: Hm I'm not a fan of adding these fields to `slurp_box`.

Maybe we can do the computation in `print_formatted_result` instead? Would require to change the arg to something like `slurp_selection` instead of `slurp_box`.

emersion:
Superseded by https://github.com/emersion/slurp/pull/111

Issue #102, opened 2022-02-06
`\n` in format-string doesn't print newline


subnut:
The manpage shows that -
```
The default format is "%x,%y %wx%h\n"
```
which contains a trailing newline that corresponds the the newline at the end of the output..

But, if we run -
```
slurp -f "%x %y\n"
```
It prints -
```
1220 642\n
```
**WITHOUT** a trailing newline. (The `\n` is printed as-is)

---

Since there is no trailing newline, piping the result into the _shell builtin_ `read` doesn't work.

tmccombs:
slurp doesn't handle backslash escapes itself, you'll need to escape it at the shell level. For example in bash or zsh you would do:

```
slurp -f $'%x %y\n'
```

I'm going to go ahead and close this issue, but if you think there is something that still needs to be done, we can re-open it.

subnut:
Okay, makes sense. But shouldn't it be mentioned somewhere in the manpage?

subnut:
@tmccombs This doesn't work -
```sh
slurp -f "%x %y$(printf '\n')"
```

tmccombs:
That's because `$(printf '\n')` strips the trailing newline.

tmccombs:
> But shouldn't it be mentioned somewhere in the manpage?

Made a PR to add it to the manpage.

PR #103, opened 2022-02-08
Add documentation of how to use backslash escapes.

tmccombs:
See #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,20 @@ Select a window under Sway, using `swaymsg` and `jq`:
 swaymsg -t get_tree | jq -r '.. | select(.pid? and .visible?) | .rect | "\(.x),\(.y) \(.width)x\(.height)"' | slurp
 ```
 
+
 Select a window without border under Sway, using `swaymsg` and `jq`:
 
 ```sh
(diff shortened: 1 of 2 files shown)


emersion:
TBH this seems out of place for the man page.

tmccombs:
I agree with @subnut that the current text is a little misleading on how slurp handles \n, and it would probably be good to direct users on how to add newlines to a custom format string. 

Do you have a better solution?

I suppose some other options are to always add a newline to the end of the output or to add a "%n" escape code to add a newline.

columbarius:
Alternative: Adding this at least to the README.md file, which already contains some examples?

emersion:
My guess is that it doesn't really matter whether there's a newline or not, `read(1)` and `getline(3)` should still behave properly when there's no final newline. The fact that the default format includes a final newline is just a nicety to have slightly better output when the user manually invokes slurp (avoid having the shell prompt on the same line).

slurp's man page is not a good place to explain how to use a shell, especially when it's Bash specific.

I'd just change the man page to e.g.

> The default format is "%x,%y %wx%h" followed by a newline.

tmccombs:
I've removed the shell specific note from the man page, and instead added an example to the README.

emersion:
This is also missing a review.

PR #109, opened 2022-08-13
Add option to display labels

columbarius:
This introduces the option flags `P, l, L` which can be used to display a box with the of the selectable rectangle inside.

`l, L` are used for the text and background colours.
`P` is used for the position to anchor the labelbox. The same syntax is used as in wlr_layer_surface with an additional 'center' keyword.


main.c
```diff
@@ -717,6 +722,26 @@ uint32_t parse_color(const char *color) {
 	return res;
 }
 
+enum slurp_label_anchor parse_label_anchor(const char *anchor) {
```
tmccombs, line 725: "center-top" is equivalent to "top-center" is equivalent to "center" is equivalent to "top-bottom-left-right-center". 

It will also happily accept something like "top-nonsenes" and treat it the same as "top" or "not-top" is also the same as "top", or "neotop". 

Should this be more strict in what it accepts?

render.c
```diff
@@ -19,6 +22,45 @@ static void draw_rect(cairo_t *cairo, struct slurp_box *box, uint32_t color) {
...
+	struct slurp_box labelbox = { 0 };
+	labelbox.width = ceil(extents.width) + 2*LABEL_BOX_PADDING;
+	labelbox.height = ceil(extents.height) + 2*LABEL_BOX_PADDING;
+	labelbox.x = box->x + (box->width - labelbox.width) / 2;
+	labelbox.y = box->y + (box->height - labelbox.height) / 2;
+
+	if (anchor & ANCHOR_TOP && anchor & ANCHOR_BOTTOM) {
+		// Nothing to do
+	} else if (anchor & ANCHOR_TOP) {
+		labelbox.y = box->y + 20;
```
tmccombs, line 44: The 20 and 10 should probably be factored out as LABEL_VERTICAL_MARGIN and LABEL_HORIZONTAL_MARGIN or similar.

main.c
tmccombs: Also "top-bottom" is the same as "center" (and similarly "left-right")

render.c
```diff
@@ -19,6 +22,45 @@ static void draw_rect(cairo_t *cairo, struct slurp_box *box, uint32_t color) {
...
+	} else if (anchor & ANCHOR_BOTTOM) {
+		labelbox.y = box->y + box->height - 20 - labelbox.height;
+	}
+	if (anchor & ANCHOR_LEFT && anchor & ANCHOR_RIGHT) {
+		// Nothing to do
+	} else if (anchor & ANCHOR_LEFT) {
+		labelbox.x = box->x + 10;
```
tmccombs, line 51: why is the horizonal margin different than the vertical margin?

tmccombs (changes requested):

render.c
columbarius: Done

render.c
columbarius: I've used the same offset as used for the dimensions.

main.c
columbarius: I thought this lack of strictness simplifies parsing without any harm/conflict. 
The structure of the anchor is the same as used for `wlr_layer_surface` and I just added center as convenience. As such the same duplications happen, but I don't se an issue there.

PR #110, opened 2022-08-14
Print \n in format string correctly

columbarius:
Previously \n from a format string was read as two seperate characters
instead of a newline.

Before:

screenshot-2022-08-14-15:20:23
(Please ignore the -, it's already fixed)

After:

screenshot-2022-08-14-15:19:47

### DIFF
--- a/main.c
+++ b/main.c
@@ -768,6 +768,19 @@ static void print_formatted_result(FILE *stream, const struct slurp_box *result,
 				i--;
 			}
 		}
+		if (c == '\\') {
+			char next = format[i + 1];
+			i++; // Skip the next character
+			switch (next) {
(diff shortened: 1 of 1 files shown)


tmccombs:
Have you seen #102?

If we were to add an escape code for newline directly to slurp, I would prefer to use "%n" so that we  don't have two different escape prefixes.

columbarius:
No, haven't seen that. Thanks for linking. Haven't seen this documented in the man page yet (and the default string with "format" is a little bit misleading imho). Should I add this? Didn't know that `$'string'` would replace escape chars.

columbarius:
Oh sorry. Haven't seen (and looked) for it.

PR #111, opened 2022-08-14
Print output local coordinates

columbarius:
With the '-x' flag slurp will calculate the x and y coordinates relative
to the position of the output the selection starts on (upper left corner).
If the selection spans over multiple outputs it will be cropped to the
intersection with the output.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1,5 +1,6 @@
 #define _POSIX_C_SOURCE 200809L
 
+#include <assert.h>
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -44,6 +45,10 @@ static int32_t box_size(const struct slurp_box *box) {
 
 static int max(int a, int b) {
 	return (a > b) ? a : b;
+}
+
+static int min(int a, int b) {
+	return (a < b) ? a : b;
 }
 
 static struct slurp_output *output_from_surface(struct slurp_state *state,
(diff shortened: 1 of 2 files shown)


emersion:
I'd rather not add a global flag if we are to add such a feature. I'd prefer new `%` sequences.

columbarius:
Are `%X, %Y` fine?

columbarius:
gentle ping

main.c
```diff
@@ -1025,7 +1055,7 @@ int main(int argc, char *argv[]) {
 		fprintf(stderr, "selection cancelled\n");
 		status = EXIT_FAILURE;
 	} else {
-		print_formatted_result(stream, &state.result, &state.outputs, format);
+		print_formatted_result(stream, &state, &state.result, format);
```
emersion, line 1058: If we're passing the state, no need to pass the result separately.

slurp.1.scd
```diff
@@ -89,6 +89,18 @@ Interpreted sequences are:
 
 %h	The height of the selection
 
+%X  The x-coordinate of the selection with coordinates relative to the output
```
emersion, line 92: Nit: should use a tab for proper alignment.

slurp.1.scd
```diff
@@ -89,6 +89,18 @@ Interpreted sequences are:
...
+	containing the top left corner.
+
+%Y  The y-coordinate of the selection with coordinates relative to the output
```
emersion, line 95: Ditto

emersion (changes requested):
Overall looks good, just a few finishing touches

emersion (approved):
LGTM, thanks!

WhyNotHugo:
Can we have a new release with this change?

PR #113, opened 2022-09-04
Implement option to use image as backgrop

tmccombs:
So that you can select a region of a full screenshot from grim (or
similar).

I'm not sure if this will work with rotated or flipped displays yet.

### DIFF
--- a/include/pool-buffer.h
+++ b/include/pool-buffer.h
@@ -16,6 +16,10 @@ struct pool_buffer {
 	bool busy;
 };
 
+struct pool_buffer *create_buffer(struct wl_shm *shm,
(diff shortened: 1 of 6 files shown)


deifactor:
Cleaner version:

```shell
#!/bin/sh

set -euo pipefail
grim -o $(swaymsg -t get_outputs | jq -r '.[] | select(.focused) | .name') - |
    swayimg -s fit -f -b none - &
swayimg_pid=$!
grimshot $@
kill $swayimg_pid
```

Invoke it the same way you'd invoke grimshot.

emersion:
Sorry, but I don't think this feature is desirable.

PR #126, opened 2023-04-07
Distinguish cancellation by esc from click

cjacker:
I have a requirement to distinguish cancellation by escape from click, maybe somebody also need this?

My setup of grim and slurp:
- select to take a screenshot of area.
- click to take a screenshot of whole screen.
- escape to cancel

Currently, both click and escape return 'selection cancelled', this is a try to distinguish these two kind of cancellation. After patched, click will return '0,0 0x0', and escape will return 'selection cancelled'.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1054,8 +1054,12 @@ int main(int argc, char *argv[]) {
 		// This space intentionally left blank
 	}
 
-	if (state.result.width == -1 && state.result.height == -1) {
+	if (state.result.width == 0 && state.result.height == 0) {
 		fprintf(stderr, "selection cancelled\n");
+		status = EXIT_FAILURE;
+	} else if (state.result.width == -1 && state.result.height == -1) {
+		// print a different msg if cancelled by ESC
+		fprintf(stderr, "selection cancelled by ESC\n");
 		status = EXIT_FAILURE;
 	} else {
 		print_formatted_result(stream, &state, format);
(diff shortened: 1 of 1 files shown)


emersion:
Can you pass a region to stdin which covers the whole screen instead?

cjacker:
> Can you pass a region to stdin which covers the whole screen instead?

I bind below script to Print key, After 'print' pressed, both click and escape print 'selection cancelled'. If I want to take a whole screenshot, without patch I have to bind it to another key, I intend to use 'escape` to cancel and 'click' to take whole screenshot, this is the patch intend to do. 

If  printing out '0,0 0x0' can not be accepted,  click to return 'selection cancelled' and `esc` to return a different msg string such as 'selection cancelled by esc' is also OK for this Scenario.

```bash
#!/usr/bin/bash

IMG_VIEWER=true
for cmd in pqiv swayimg; do
  if command -v "$cmd" > /dev/null 2>&1; then
    IMG_VIEWER=$cmd
    break
  fi
done

FILENAME=screenshot-$(LANG=C LC_ALL=C date "+%Y-%m-%d-%H-%M-%S").png

# click will print "0,0 0x0"
# escape pressed will print "selection cancelled"
SELECT=$(LANG=en_US slurp 2>&1)

# if slurp cancelled, grim will not take screenshot
if [ "$SELECT""x" == "selection cancelled""x" ]; then
  true
elif [ "$SELECT""x" == "0,0 0x0""x" ]; then
  grim "$HOME"/"$FILENAME" && $IMG_VIEWER "$HOME"/"$FILENAME"
else
  grim -g "$SELECT" "$HOME"/"$FILENAME" && $IMG_VIEWER "$HOME"/"$FILENAME"
fi
```

cjacker:
Update, it print a different msg now if cancelled by ESC.

PR #127, opened 2023-04-07
Add a debug msg to distinguish cancellation by ESC and click

cjacker:
As mentioned in https://github.com/emersion/slurp/pull/126, I bind a script to 'Print' key and intend to use 'ESC' to cancel and 'click' to take a whole screenshot. 

Currently, both ESC and click print out same msg  'selection cancelled', and I need a different msg to distinguish cancellation by ESC from click.

The previous try in https://github.com/emersion/slurp/pull/126 is stupid. What I need is just a different msg, one more `fprintf` is enough for that.

The script I use:
```
#!/usr/bin/bash

IMG_VIEWER=pqiv

FILENAME=screenshot-$(LANG=C LC_ALL=C date "+%Y-%m-%d-%H-%M-%S").png

SELECT=$(LANG=en_US slurp 2>&1)

if [ "$SELECT""x" == "ESC pressed selection cancelled""x" ]; then
  true
elif [ "$SELECT""x" == "selection cancelled""x" ]; then
  grim "$HOME"/"$FILENAME" && $IMG_VIEWER "$HOME"/"$FILENAME"
else
  grim -g "$SELECT" "$HOME"/"$FILENAME" && $IMG_VIEWER "$HOME"/"$FILENAME"
fi
```

### DIFF
--- a/main.c
+++ b/main.c
@@ -212,8 +212,8 @@ static void handle_selection_start(struct slurp_seat *seat,
 			state->running = false;
 		}
 	} else {
-		current_selection->anchor_x = current_selection->x;
-		current_selection->anchor_y = current_selection->y;
+		state->result.x = current_selection->anchor_x = current_selection->x;
+		state->result.y = current_selection->anchor_y = current_selection->y;
 	}
 }
 
@@ -306,6 +306,8 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
 			seat->touch_selection.has_selection = false;
 			state->edit_anchor = false;
 			state->running = false;
+			state->result.x = -1;
+			state->result.y = -1;
 			break;
 
 		case XKB_KEY_space:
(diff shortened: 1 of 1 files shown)


emersion:
Error messages should not be relied upon. They might change, be localized, etc.

cjacker:
> Error messages should not be relied upon. They might change, be localized, etc. They are for human consumption only.

I am not sure which is your prefer way, and took a look at hacksaw, 'click' print 'global cursor pos 0x0' and 'ESC' report a cancellation msg, it can be used to:
- distinguish 'click' from 'ESC'.
- take a screenshot of window at global cursor pos if screenshot utility support it.

tmccombs:
Would a different status code fit your needs?

tmccombs:
I kind of like the idea of a single click just creating a 0x0 selection. However, I'm a little concerned that that might break things where the result is sent to something that doesn't expect a selection of zero size.  That could be mitigated by gating this behavior with an option (maybe "-P" for "allow point"), but that feels a little weird to too. Maybe add an option for now, but mention in the man page that that behavior may change to the default in the future?

PR #129, opened 2023-05-25
Makes a single click equivalent to a point selection

fuyarovh:
This is useful to create a maim like screenshot utility where a single click can select a window while a drag action selects a rectangle. In order to still have a way to cancel slurp with the mouse, buttons besides the left key act like escape.

### DIFF
--- a/main.c
+++ b/main.c
@@ -222,8 +222,20 @@ static void handle_selection_end(struct slurp_seat *seat,
 	}
 	if (current_selection->has_selection) {
 		state->result = current_selection->selection;
+	} else {
+		state->result.x = current_selection->x;
+		state->result.y = current_selection->y;
+		state->result.width = state->result.height = 1;
 	}
 	state->resizing_selection = false;
+	state->running = false;
+}
+
+static void handle_selection_cancelled(struct slurp_seat *seat) {
+	struct slurp_state *state = seat->state;
+	seat->pointer_selection.has_selection = false;
+	seat->touch_selection.has_selection = false;
+	state->edit_anchor = false;
 	state->running = false;
 }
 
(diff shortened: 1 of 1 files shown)


emersion:
Does the -p option do what you want?

fobekyo:
The p option does not achieve the same goal because it requires to specify in advance if I want to have a point or an area selection

pacien (changes requested):
I've hit the same need to obtain either a rectangular or point selection from the same call of `slurp` for the same use case (selecting a rectangle or a window for screenshotting).

This is distinct from `-p` which prevents the selection of a rectangle in the UI. Here we want both possibilities with the same call.

The responsibility to deal with either a `1x1` or bigger selection is left to the caller when both are enabled. Nothing hard.

Regarding the right click action to cancel selection: please mention it in the documenation (manual page).

main.c
```diff
@@ -238,14 +246,19 @@ static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
...
 	seat->button_state = button_state;
-
-	switch (button_state) {
-	case WL_POINTER_BUTTON_STATE_PRESSED:
-		handle_selection_start(seat, &seat->pointer_selection);
-		break;
-	case WL_POINTER_BUTTON_STATE_RELEASED:
-		handle_selection_end(seat, &seat->pointer_selection);
+	switch (button) {
+		case 272: //left mouse button
```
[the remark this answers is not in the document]
Funestia: whats wrong about the nested switch in this instance?

b0o:
This would be really nice, +1. My use-case is, in Niri, to allow the user to either pick a window by point-clicking, or a specific bounding box by click-dragging. I've tested this and it does exactly what I want. If there's concern about breaking backwards compatibility, perhaps this behavior could be moved behind a flag.

main.c
```diff
@@ -236,13 +248,19 @@ static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
...
+			handle_selection_start(seat, &seat->pointer_selection);
+			break;
+		case WL_POINTER_BUTTON_STATE_RELEASED:
+			handle_selection_end(seat, &seat->pointer_selection);
+			break;
+		}
 		break;
-	case WL_POINTER_BUTTON_STATE_RELEASED:
-		handle_selection_end(seat, &seat->pointer_selection);
+	default: //right mouse button
```
tmccombs: This could also be a middle mouse button, or one of the forward/backward mouse buttons, etc. 

Which I think is probably fine implementation-wise, but the comment should at least be changed to reflect that. 

I'll also note that currently, it is possible to select with right click instead of with left click, but I don't think that is intentional.

tmccombs (changes requested):
One small thing that should be improved but otherwise looks good.

main.c
fuyarovh: Before this change you could cancel a selection with a single click. Because this is no longer possible with this change I changed other mouse buttons to mean cancel but I can change that if that behaviour is not wanted.

tmccombs (approved):

PR #132, opened 2023-06-25
Enable Shift key only while resizing current selection

### LINKED ISSUE #131: Disable Shift key when current selection is a predefined rectangle
Pressing Shift while hovering over a predefined rectangle results in weird behavior where the anchor position is taken to be the top left of the screen (the point [0, 0]), and the cursor position defines the other corner. Moving the cursor re-highlights the predefined rectangle again.

One idea for a fix is to allow Shift only while the pointer button is pressed/moving. In other words, the current selection is guaranteed not to be a predefined rectangle.

mkrasnitski:
Pressing Shift still changes the aspect ratio, but that is only reflected while resizing a selection (not while hovering). 

Closes #131

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -54,6 +54,7 @@ struct slurp_state {
 	bool display_dimensions;
 	bool single_point;
 	bool restrict_selection;
+	bool resizing_selection;
 	struct wl_list boxes; // slurp_box::link
 	bool fixed_aspect_ratio;
 	double aspect_ratio;  // h / w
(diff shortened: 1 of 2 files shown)


MonsterDruide1:
I'm experiencing the same bug and can confirm that this PR solves the issue.

It is particularly annoying when using a hotkey to launch `slurp` (as part of a larger construct to take partial screenshots, in my case `grimshot`) which includes Shift, which means that this key is already pressed when `slurp` is starting up. Given that the entire screen is covered in pre-configured selections, the selection always jumps to the top-left corner.

@tmccombs seems to be the only active contributor at the moment, any chance this fix can be merged soon?

PR #134, opened 2023-07-29
Implement cursor_shape

Nefsen402:


protocol/meson.build
```diff
@@ -17,6 +17,8 @@ wayland_scanner_client = generator(
 
 client_protocols = [
 	wl_protocol_dir / 'stable/xdg-shell/xdg-shell.xml',
+	wl_protocol_dir / 'staging/cursor-shape/cursor-shape-v1.xml',
+	wl_protocol_dir / 'unstable/tablet/tablet-unstable-v2.xml',
```
emersion, line 21: Nit: we don't need this protocol

Nefsen402: A linking error occurs without it.

emersion: Ah, right. Probably an access to the `zwp_tablet_tool_v2_interface` global variable.

emersion (approved):
Thanks!

PR #138, opened 2023-08-15
Add option flag to support re-selecting until Return key is pressed

m-bartlett:
Adds `-C` which prevents slurp from exiting immediately once the rectangle drag is released. The rectangle may be re-selected, and selection is ended by pressing the Return key. The intent of this feature is to enable users to redo a selection if the first attempt was unsatisfactory.

slurp


heroin-moose:
Bump? This is a life safer for me.

markg85:
Ohh there's a PR for this!! Awesome! But why the holdup? 2023.. what's missing before it can be merged?

main.c
```diff
@@ -185,10 +185,6 @@ static void pointer_handle_leave(void *data, struct wl_pointer *wl_pointer,
 static void pointer_handle_motion(void *data, struct wl_pointer *wl_pointer,
 		uint32_t time, wl_fixed_t surface_x, wl_fixed_t surface_y) {
 	struct slurp_seat *seat = data;
-	// the places the cursor moved away from are also dirty
-	if (seat->pointer_selection.has_selection) {
```
tmccombs, line 189: Why is this removed?

main.c
```diff
@@ -324,24 +322,34 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
...
+		
 		case XKB_KEY_Shift_L:
 		case XKB_KEY_Shift_R:
 			if (!state->fixed_aspect_ratio) {
 				state->aspect_ratio = 1;
 				recompute_selection(seat);
 			}
 			break;
+
+		case XKB_KEY_Return:
```
tmccombs, line 334: What if return is hit while you are currently dragging a selection? Should we then store the current selection in the result?

main.c
```diff
@@ -324,24 +322,34 @@ static void keyboard_handle_key(void *data, struct wl_keyboard *wl_keyboard,
...
+			state->running = false;
+			break;
 		}
 		break;
 
 	case WL_KEYBOARD_KEY_STATE_RELEASED:
-		if (keysym == XKB_KEY_space) {
+		switch (keysym) {
```
tmccombs, line 341: This isn't necessarily a bad change, but it seems unrelated to the rest of this PR

main.c
```diff
@@ -713,6 +721,7 @@ static const char usage[] =
 	"  -o           Select a display output.\n"
 	"  -p           Select a single point.\n"
 	"  -r           Restrict selection to predefined boxes.\n"
+	"  -C           Reselect until enter key is pressed.\n"
```
tmccombs, line 724:
```suggestion
	"  -C           Wait until enter key is pressed.\n"
```

tmccombs (changes requested):

main.c
m-bartlett: Setting the output dirty would cause the current selection rectangle to clear whenever the mouse was moved while not dragging, here is a video of the behavior before this block was removed:

https://github.com/user-attachments/assets/a31c30ec-9b4e-4b2c-8002-b82e310696f1

main.c
m-bartlett: It lays the groundwork for some other PRs I am hoping to make, so the switch statement would get more utilized in the future for other keyboard events.

main.c
m-bartlett: The current behavior returns the selection region of the drag at the moment enter is pressed, in my opinion that's intuitive since releasing the drag can cause a slight undesired movement when trying to be precise

main.c
Samq64: I am having this issue on https://github.com/emersion/slurp/pull/138/commits/72bc7c0a568a7a140e09954be2c84fa1be275d7b.

Spider1x0:
Any updates on this? It would be a life saving feature tbh

PR #139, opened 2023-08-20
Add a flag to allow the user to alter the selection after releasing the mouse button

### LINKED ISSUE #136: Allow altering of selection after releasing mouse button
I don't know if this is possible with the way slurp is currently implemented, but it would be nice if you could add a flag that lets the user alter the selection after the initial selection has been made and the mouse button has been released similar to how flameshot does it.
UI wise this could be implemented by adding small circles or rectangles to the edges of the selection box and when dragging these the selection gets altered. When pressing the return key or some other key which can be configured the selection is finished and output to stdout.

PucklaJ:
Fixes #136

recording-2023-08-20-19:55:46

This PR adds a new flag `-A` which enables a new mode that lets the user alter the selection similar to [flameshot](https://github.com/flameshot-org/flameshot).

The following flags have been added:
+ `-g` Sets the background color of the grabbers the are used to drag the selection
+ `-G` Sets the border color of the grabbers
+ `-R` Sets the radius of the grabbers
+ `-A` Enables the new selection altering mode

Unfortunately since I don't have a touch screen to test the touch input I don't know how well this implementation works with touch. But for the most part it should work.

Here is an overview about how this was implemented:
The most important functions are `handle_selection_start`, `handle_selection_end` and `handle_alter_selection_motion`.

+ `handle_selection_start`
This functions acts like a click event. In this function it is checked if the user clicks on one of the grabbers, in the middle of the selection or somewhere else. To check if the user clicks on one of the grabbers the `circle_intersect` function is utilized. If no selection has been made yet then it basically acts the same as before (even the same functions are called). After clicking on a grabber a state machine comes into play:

  + `ALTER_STATE_NONE`
The mouse button is not pressed or no selection has been made yet.
  + `ALTER_STATE_INITIAL`
The initial selection is being made. In this state it works as usual.
  + `ALTER_STATE_TOP_LEFT` and similar states
These states occur if the user clicks on one of the grabbers. Every grabber has its own state. Depending on the state the selection is altered in a different way. Also if the user clicks on a grabber an offset (difference between mouse position and center of grabber) is calculated so that the grabber does not snap to the cursor when moving the mouse.

+ `handle_selection_end`
This functions acts like a click release event. Not a lot of changes have been made there.

+ `handle_alter_selection_motion`
This is the callback that handles mouse motion. Here depending on the state the selection is altered. After the altering the selection it is checked if the user moved the grabber over the other side (width < 0 or height < 0). If this is the case the state is swapped to the other side. Then fixed aspect ratio is handled.

To display the grabbers the `draw_grabbers` function has been added and is called in `render`.

I also changed it so that every motion alters the result. This was made to fix an issue where pressing Escape would not cancel the selection and to also fix an issue where pressing Return while still holding down the mouse button would not output the correct geometry.


heroin-moose:
Nice!

heroin-moose:
However, move is buggy:

- Select a region (region 1)
- Move the rectangle to another position (region 2)
- Press Enter

Coordinates of region 1 will be reported.

heroin-moose:
Also touch screen does not work :(

PucklaJ:
> Also on my laptop's touch screen I can't do anything: resize, move, etc. No reaction at all :(

Yes. Unfortunately I do not have a touch device where I can test it, so I couldn't implement it. For now it is mouse only. I guess someone else can come in and add support?

heroin-moose:
@PucklaJ, have you, by any chance, been able to look into the moving issue? Aside from that this PR works great for me.

PucklaJ:
@heroin-moose I have been working on [my own selection and screenshot tool](https://github.com/PucklaJ/samurai-select), because there is not much activity on this project. I aim to have feature parity with slurp and I want to do things a little bit differently and add some features which are not currently in slurp. For me (and maybe also for you) samurai-select (or smel) works just as fine if not better than slurp. If you want to try it out you just need to install go and build (`go build`) or install (`go install`) it.

PR #141, opened 2023-08-31
Fix slurp -o "selection cancelled" with touch inputs

stacyharper:
fixes: https://github.com/emersion/slurp/issues/140


emersion:
Hm it seems we call this function on pointer release and on touch down… Shouldn't we call it on touch up instead?

stacyharper:
We could move the call to `seat_update_selection` to `touch_handle_up`, and conditioned it to be run if there is still no selection at this point. It allow to select rects on simple touches, while still allowing manual box selections.

But the problem is the case of `slurp -o -r`. To handle that, we also have to add this function to `touch_handle_down`, before `handle_selection_start`. Because in this situation, `handle_selection_start` will finish the loop. Then, the `seat_update_selection` call from `touch_handle_up` becomes redundant.

emersion:
Sorry, but I still don't understand what's peculiar about touch input that makes it different from pointer input here.

stacyharper:
I don't get it. We update the selectioned box on touch down because this is the only moment we can, and have to do so.

emersion:
The pointer codepath does not call `seat_update_selection()` when a button is pressed. Why does touch down need to call it?

stacyharper:
Because we know where the pointer is before the user press any button. We don't know where the user finger is before it actually touch the screen.

stacyharper:
Gentle bump. Could this be reviewed? I continue to think the fix is good, even if it looks wrong with a first glance.

Issue #57, opened 2020-06-10
Crosshair cursor is not scaled on 4k display


nostdm:
It seems that when the selection cursor appears (crosshair) it's half the size of a normal cursor on a 4k display. Could this be a problem with the cursor theme? I tried elementary and adwaita cursor themes with the same result. How does slurp determine the size/theme of the crosshair cursor?

ferrreo:
This kinda makes slurp unusable at higher resolutions, it is so small I can't see it 90% of the time, I'll have a go at fixing it

BANanaD3V:
Any updates?

PR #144, opened 2023-10-30
Fix crosshair cursor is not scaled

colorsakura:
Fixes: #57
test on sway-git and hyprland-git, it works well.
But I don't know what these changes will destroy.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1015,10 +1015,6 @@ int main(int argc, char *argv[]) {
 		return EXIT_FAILURE;
 	}
 
-	if (!state.cursor_shape_manager && !create_cursors(&state)) {
-		return EXIT_FAILURE;
-	}
-
 	struct slurp_output *output;
 	wl_list_for_each(output, &state.outputs, link) {
 		output->surface = wl_compositor_create_surface(state.compositor);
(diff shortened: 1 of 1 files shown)


tmccombs (approved):

main.c
```diff
@@ -827,7 +827,7 @@ static void add_choice_box(struct slurp_state *state,
 	wl_list_insert(state->boxes.prev, &b->link);
 }
 
-static bool create_cursors(struct slurp_state *state) {
+static bool create_cursors(struct slurp_state *state, struct slurp_output *output) {
```
emersion, line 830: This new parameter is not used -- it's overwritten by `wl_list_for_each()`.

main.c
```diff
@@ -1067,6 +1062,10 @@ int main(int argc, char *argv[]) {
 			wl_compositor_create_surface(state.compositor);
 	}
 
+	if (!state.cursor_shape_manager && !create_cursors(&state, output)) {
```
emersion, line 1065: Moving this right after the second `wl_display_roundtrip()` should be enough: the `wl_output` properties should've been received by then.

emersion:
Thanks!

PR #149, opened 2023-12-29
Mouse release

### LINKED ISSUE #146: "selection cancelled" when I trigger slurp by clicking on a button with my mouse
### Little Context...
* I've made a **button**(_here's the problem_) in my Waybar which on-click supposed to take screenshot using `slurp` and `grim`
* I've a **key binding**(_no problem here_) in my Hyprland which also takes screenshot using `slurp` and `grim`

### What I faced?
When I use the key binding(i.e. using keyboard to trigger `slurp` and using mouse to drag) to take the screenshot, it works perfectly fine BUT  
when I click(using the mouse obviously) on the button on Waybar, it just flashes and disappear. I checked the console and it gave an output `selection cancelled` HOWEVER  
I fixed the issue by triggering `slurp` with **"on-click-release"** instead of **"on-click"**

### What I think is the problem?
In my guess `slurp` goes through 3 actions,
* Triggering slurp command
* Mouse click to start dragging
* Mouse release to end dragging

Here, when I click the Screenshot-button(_triggered 1st action_), I already got my mouse in clicked state after that I release(_triggered 3rd action_) my mouse(thinking to start dragging), it just disappears and sends the `selection cancelled` as stdout because the last action has already been triggered without the 2nd action.

### Possible fix
Don't cancel the selection on mouse-release if mouse-click hasn't been triggered yet, instead allow it to start dragging again.

tmccombs:
Fixes #146


emersion:
This sounds like a compositor bug, rather than a slurp bug.

PR #151, opened 2024-03-16
fix handling cursor_shape_manager

apprehensions:
previously, cursor_shape_manager was not handled and registered, which can lead to slurp incorrectly loading the entire cursor theme, despite the compositor supporting cursor_shape_manager.


emersion (approved):
Thanks!

PR #158, opened 2024-07-10
Fix weird `:;` code style

xnuk:


main.c
```diff
@@ -273,7 +273,7 @@ static void keyboard_handle_keymap(void *data, struct wl_keyboard *wl_keyboard,
 	case WL_KEYBOARD_KEYMAP_FORMAT_NO_KEYMAP:
 		seat->xkb_keymap = xkb_keymap_new_from_names(seat->state->xkb_context, NULL, XKB_KEYMAP_COMPILE_NO_FLAGS);
 		break;
-	case WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1:;
```
emersion, line 276: Unfortunately we need to keep this one because of a C language quirk: variable declarations can't follow a `case`.

emersion:
We can remove the other two as you did though, because the declarations originally there are now gone!

xnuk:
Is this a common C code style? I'm still unfamiliar with C (for years).

emersion:
It's one way to workaround the annoyance in the C language. Another way is to use brackets:

```c
switch (value) {
case FOO: {
    do_something();
    break;
}
}
```

But it makes the indentation look funny so I prefer `:;`.

emersion:
Thanks!

PR #161, opened 2024-12-03
Add option to cancel selection on any keypress

mrsobakin:
Useful when `slurp` is binded to some hotkey (e.g. Print, for taking screenshots). Prevents multiple `slurp'`s from spawning when the key is held.

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -59,6 +59,7 @@ struct slurp_state {
 	struct wl_list boxes; // slurp_box::link
 	bool fixed_aspect_ratio;
 	double aspect_ratio;  // h / w
+	bool cancel_on_keypress;
 
 	struct slurp_box result;
 };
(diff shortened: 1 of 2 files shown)


emersion:
This sounds like a workaround. The proper fix would be to disable key repeat for the key binding.

mrsobakin:
Even if key repeat is disabled, you can still create multiple instances of `slurp`  by pressing the hotkey multiple times, which is not the intended behavior for screenshot utility (or any utility that involves selecting screen region, for that matter).

tmccombs:
I've run into this with multiple key presses. Another option is to only allow a single slurp process to run per session, or maybe per seat.

mrsobakin:
I actually like your idea of limiting slurp instances per session more; however I'm not sure on how we should approach implementing it. Does Wayland support storing global user data on the server side? Or should we just keep track of slurp instances via files in `XDG_RUNTIME_DIR`, à la pid files?

tmccombs:
I'd probably implement it as a lock file in $XDG_RUNTIME_DIR that includes either the value of `$WAYLAND_DISPLAY` or the id of the seat in the name.

mrsobakin:
Superseded by #162.

PR #162, opened 2024-12-04
Add pidfiles to prevent multiple instances from running

mrsobakin:
Prevents multiple slurp instances from running on the same display at the same time.

Can be disabled via `-m` flag, allowing multiple instances.

Supersedes #161.


tmccombs (changes requested):
Use flock to prevent races.

tmccombs:
Superceded by #174

PR #175, opened 2025-09-10
Enforce format with clang-format

tmccombs:
I tried to use config that minimized the changes, but there were some
    existing inconsistencies in the format and some things I couldn't quite
    get clang-format to due.
    
    For example, I couldn't find a way to get clang-format to use more
    indentation for continuation of parameter lists than for other types of
    line continuation.
    
    I also used ColumnLimit: 0, because anything else would add considerable
    changes to either put things on the same line, split lines up, or both.
    
    I figured if we want to set a limit, we could do that in a separate PR.
    
    I'm also totally open to changing any of the configuration options.
    
    If clang-format is too prescriptive, I could also just add something to
    enforce that indentation is done with tabs instead of spaces.

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -5,6 +5,8 @@ packages:
   - wayland-protocols
   - cairo
   - libxkbcommon
+  # needed for clang-format
+  - clang
 sources:
   - https://github.com/emerison/slurp
 tasks:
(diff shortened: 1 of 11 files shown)


emersion:
clang-format cannot enforce the style used by wlroots (and slurp).

tmccombs:
Is there a rool that can? Currently, there is some inconsistency in style.

PR #176, opened 2025-09-20
Add mirror mode with Control key

axel-angel:
This implements a new mirror behavior while pressing Ctrl: the box selection will grow outwards from the starting selection (first click) symmetrically. This also works in conjunction with restricted aspect ratio (-a flag or shift key) and/or edit anchor (space key). The doc is also updated to mention this feature.

Demo follows.

https://github.com/user-attachments/assets/8fa0c7e3-2877-43fe-a20b-41727f7241f1

The code isn't very straightforward, there may be a better way to implement this. ~~Also the behavior is a bit surprising when switching between regular and mirror mode~~, any idea is welcomed to do this differently.

*EDIT*: demo updated with last change.

### DIFF
--- a/main.c
+++ b/main.c
@@ -47,6 +47,8 @@ static void move_seat(struct slurp_seat *seat, wl_fixed_t surface_x,
 	if (seat->state->edit_anchor) {
 		current_selection->anchor_x += x - current_selection->x;
 		current_selection->anchor_y += y - current_selection->y;
+		current_selection->start_x += x - current_selection->x;
+		current_selection->start_y += y - current_selection->y;
 	}
 
 	current_selection->x = x;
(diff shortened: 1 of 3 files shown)


main.c
```diff
@@ -97,17 +99,36 @@ static void handle_active_selection_motion(struct slurp_seat *seat, struct slurp
...
-	int32_t width = abs(dist_x) + 1;
-	int32_t height = abs(dist_y) + 1;
-	if (seat->state->aspect_ratio) {
+
+	int32_t width;
+	int32_t height;
+
+    if (seat->state->mirror_mode) {
+        width  = 2 * abs(start_x - current_selection->x);
+        height = 2 * abs(start_y - current_selection->y);
```
axel-angel, line 112: Should we also include the +1 offset implemented below?

main.c
```diff
@@ -97,17 +99,36 @@ static void handle_active_selection_motion(struct slurp_seat *seat, struct slurp
...
-	current_selection->selection.y = dist_y > 0 ? anchor_y : anchor_y - (height - 1);
+    if (seat->state->mirror_mode) {
+        int32_t delta_x = width / 2;
+        int32_t delta_y = height / 2;
+        anchor_x = start_x - delta_x;
+        anchor_y = start_y - delta_y;
+    }
+
+	current_selection->selection.x = dist_x > 0 || seat->state->mirror_mode ? anchor_x : anchor_x - (width - 1);
+	current_selection->selection.y = dist_y > 0 || seat->state->mirror_mode ? anchor_y : anchor_y - (height - 1);
```
axel-angel, line 131: Again, the -1 offset when dist is negative may be required for mirror mode. We may need to structure the code a bit differently to write the code only once.

PS: is the -1 to compensate the border? in that case we need to use state border_weight instead.

axel-angel:
> Also the behavior is a bit surprising when switching between regular and mirror mode

commit https://github.com/emersion/slurp/pull/176/commits/636efee0a54ac633e7a80316362773af640dd9d3 addresss this issue making the transition predictable and flawless.

PR #183, opened 2025-12-12
feat: add `-O` to print focused output geometry immediately

mbroemme:
## Description
Take a screenshot with `grim` using `slurp` of the currently focused display with the `Print` key. Press `Print` once to grab a screenshot of the monitor your mouse is currently on (the focused output), without selecting a region or clicking.

## Why
On multi-monitor setups, "full-screen screenshot" often captures all outputs (huge image) or the wrong monitor. With `slurp -O`, the screenshot is always of the display you’re actively using.

## Example:
Bind `Print` to:
```
sh -c 'grim -g "$(slurp -O)" ~/Pictures/Screenshots/$(date +%F_%H-%M-%S).png'
```
A PNG file is created containing exactly the focused output (its x/y origin and width/height), with no interactive selection step.

## Implementation
Add a new `-O` option that queries the output currently containing the pointer and prints its box using the existing format string machinery, then exits without entering region selection. Reject `-O` when combined with `-p` or `-r` to avoid ambiguous behavior. This enables grabbing the active display dimensions instantly (e.g. for scripts) without requiring a click/drag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ Select an output and print its name:
 slurp -o -f "%o"
 ```
 
+Print the geometry of the output currently containing the pointer:
+
+```sh
+slurp -O
+```
+
 Select a window under Sway, using `swaymsg` and `jq`:
 
 ```sh
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -51,6 +51,7 @@ struct slurp_state {
   bool single_point;
   bool restrict_selection;
   bool crosshairs;
+  bool print_focused_output; // Print focused output geometry and exit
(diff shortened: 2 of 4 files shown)


emersion:
I don't think showing a layer surface and then immediately closing it is a good way to get the focused output.

I think this should be handled on the capture side instead. See e.g. https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/issues/116

mbroemme:
Which basically means to replace `wlr-screencopy` with `ext-image-copy-capture` in `grim`, correct?

PR #191, opened 2026-02-26
feat: Add rounded and chamfered border options

Tartarus6:
### New flags:
- `-R n`: Set border radius.
- `-C`:   Set border to use 45 degree chamfer instead of rounding.

### Behavior:
- If radius is not set, then the normal square rectangle will be drawn.
- If double the radius is less than the width or the height, then the radius is temporarily set to the minimum of half the height or half the width (to make sure that the radius can fit on the rectangle).
- If the width or height of the selection box is less than double the border weight, then the border is drawn as a square rectangle (to prevent weird shapes).

### Examples:
image
image

### DIFF
--- a/include/slurp.h
+++ b/include/slurp.h
@@ -47,6 +47,8 @@ struct slurp_state {
   const char *font_family;
 
   uint32_t border_weight;
+  uint32_t border_radius;
+  bool border_chamfered;
   bool display_dimensions;
   bool single_point;
   bool restrict_selection;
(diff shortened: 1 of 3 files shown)


Tartarus6:
I just found the tab in this repo that explains the contribution guidelines. I see that my commit (since there is just one) does not match the requested format.

Should I, or how should I, fix this? Or can the commit just be renamed upon merging?

